PR #34, opened 2023-08-21
bt-16: Create global styles

### LINKED ISSUE #16: Create global styles
Create global styles object with styles that will be used most often in the app (flex, flex-direction, etc.)

anogire:


mobile/src/bundles/common/enums/styles/styles.enum.ts
```diff
@@ -0,0 +1,9 @@
+const AlignmentValues = {
+    CENTER: 'center',
+    FLEX_START: 'flex-start',
+    FLEX_END: 'flex-end',
+    STRETCH: 'stretch',
+    SPACE_BETWEEN: 'space-between',
+} as const;
```
katerynakostikova:
```suggestion
const AlignmentValue = {
    CENTER: 'center',
    FLEX_START: 'flex-start',
    FLEX_END: 'flex-end',
    STRETCH: 'stretch',
    SPACE_BETWEEN: 'space-between',
} as const;
```

mobile/src/bundles/common/styles/global-styles.ts
```diff
@@ -0,0 +1,89 @@
...
+    alignSelfStretch: {
+        alignSelf: AlignmentValues.STRETCH as FlexAlignType,
+    },
+    defaultBorderRadius: {
+        borderRadius: 9,
+    },
+    defaultBorderTopRadius: {
+        borderTopRightRadius: 9,
+        borderTopLeftRadius: 9,
+    },
```
katerynakostikova: I'm not sure if it'll be used somewhere. We can add it later if needed.

mobile/src/bundles/common/styles/global-styles.ts
```diff
@@ -0,0 +1,89 @@
...
+const globalStyles = StyleSheet.create({
+    defaultMarginVertical: {
+        marginVertical: 20,
+    },
+    defaultMarginHorizontal: {
+        marginHorizontal: 20,
+    },
+    defaultMarginTop: {
+        marginTop: 20,
+    },
```
katerynakostikova: Let's add some values more values to margins and paddings, e.x. m5 = margin: 5, m10 = margin: 10, and so on till 20-25 (I guess there is no need for a bigger value). Same with margin-top/bottom/vertical etc.

mobile/src/bundles/common/styles/global-styles.ts
```diff
@@ -0,0 +1,89 @@
+import { type FlexAlignType, type FlexStyle, StyleSheet } from 'react-native';
+
+import { AlignmentValues } from '~/bundles/common/enums/enums';
+
+const globalStyles = StyleSheet.create({
```
katerynakostikova, line 8: To divide some logic, you can create some margins object, paddings object, flex object etc. and just spread it in globalStyles 
```
const globalStyles = StyleSheet.create({
	...flex,
	...margins,
	...paddings,
});
```

katerynakostikova: Also, you can apply created global styles in app.tsx and remove styles.ts from there.

mobile/src/bundles/common/enums/styles/styles.enum.ts
anogire: I transferred it to the flex.ts file. But I made constants for the values because of the error "using a value instead of a constant more than 3 times".

mobile/src/bundles/common/styles/global-styles.ts
anogire: I was looking at the Profile-page... but, yes, maybe there is a different design there

mobile/src/bundles/common/styles/global-styles.ts
anogire: I forgot about app.tsx and styles.ts for it ... I will do it after your comments

anogire: Done

mobile/src/app/app.tsx
```diff
@@ -5,15 +5,14 @@ import React, { type FC } from 'react';
 import { GestureHandlerRootView } from 'react-native-gesture-handler';
 import { Provider as StoreProvider } from 'react-redux';
 
+import { globalStyles } from '~/bundles/common/styles/global-styles';
```
katerynakostikova: Let's create styles.ts in styles folder and re-export global styles 
```suggestion
import { globalStyles } from '~/bundles/common/styles';
```

mobile/src/bundles/common/styles/flex.ts
```diff
@@ -0,0 +1,71 @@
...
+    CENTER: 'center',
+    FLEX_START: 'flex-start',
+    FLEX_END: 'flex-end',
+    STRETCH: 'stretch',
+    SPACE_BETWEEN: 'space-between',
+    SPACE_AROUND: 'space-around',
+    SPACE_EVENLY: 'space-evenly,',
+} as const;
+
+const flex = StyleSheet.create({
```
katerynakostikova: It should be just a simple object to avoid nested StyleSheet.create 
```suggestion
const FLEX_STYLES = {
```

mobile/src/bundles/common/styles/margins.ts
```diff
@@ -0,0 +1,111 @@
+import { StyleSheet } from 'react-native';
+
+const margins = StyleSheet.create({
```
katerynakostikova: same here

mobile/src/bundles/common/styles/paddings.ts
```diff
@@ -0,0 +1,111 @@
+import { StyleSheet } from 'react-native';
+
+const paddings = StyleSheet.create({
```
katerynakostikova: here also

mobile/src/bundles/common/styles/sizes.ts
```diff
@@ -0,0 +1,21 @@
+import { StyleSheet } from 'react-native';
+
+const sizes = StyleSheet.create({
```
katerynakostikova: and here

katerynakostikova (approved):

illia-dobro (approved):

nikita-remeslov (approved):

PR #49, opened 2023-08-21
bt-11: + checkbox component

### LINKED ISSUE #11: Create checkbox component
Image

https://mui.com/material-ui/react-checkbox/

ptarho:
I have created custom checkbox component with a few basic props to customize it if needed.
The example of usage:
checkbox-3
How it looks on the page:
checkbox-1
checkbox-2

When the global styles will be implemented is better to change hex colors values with corresponding data


anton-otroshchenko (approved):

frontend/src/bundles/common/components/checkbox/checkbox.tsx
```diff
@@ -0,0 +1,55 @@
+import {
+    Checkbox as CheckboxMUI,
+    FormControlLabel,
+    type SxProps,
+    type Theme,
+} from '@mui/material';
+
+type Properties = {
+    label?: string;
+    disabled?: boolean;
```
anton-otroshchenko: https://github.com/BinaryStudioAcademy/quality-criteria/blob/production/src/javascript.md

B19.

frontend/src/bundles/common/components/checkbox/checkbox.tsx
```diff
@@ -0,0 +1,55 @@
...
+    defaultChecked?: boolean;
+    checked?: boolean;
+    required?: boolean;
```
anton-otroshchenko: Same here

anton-otroshchenko (changes requested):

frontend/src/bundles/common/components/checkbox/checkbox.tsx
```diff
@@ -0,0 +1,55 @@
...
+    value?: string;
+    size?: 'small' | 'medium';
+    sx?: SxProps<Theme>;
```
nikita-remeslov: for styles we use className and modules

nikita-remeslov (changes requested):

frontend/src/bundles/common/components/checkbox/checkbox.tsx
ptarho: Can`t we just extand MUI CheckboxProps or FormControlLabelProps interface to reuse in our custom component as a props interface?
In that way we will get rid of repetetive code and make component cleaner.
The only problem is that project require us to use types, not interfaces

ptarho:
I tried to switch into styles module and create a default checkbox style but got an error with nesting depth. 
I can leave classes props as a empty string but in that way checkbox will use default MUI styles. 
The logic was that developer can pass just an array of classes to apply and than component will try to use them from local module and if there is no such classes will try to apply global
Screenshot_4
Screenshot_5

nikita-remeslov:
> I tried to switch into styles module and create a default checkbox style but got an error with nesting depth. I can leave classes props as a empty string but in that way checkbox will use default MUI styles. The logic was that developer can pass just an array of classes to apply and than component will try to use them from local module and if there is no such classes will try to apply global Screenshot_4 Screenshot_5

Can you show me an example of code, when you got the error?

frontend/src/bundles/common/components/checkbox/checkbox.tsx
```diff
@@ -0,0 +1,52 @@
...
+    isRequired = false,
+    value = '',
+    className = styles.primary,
+}) => {
+    const combinedClasses = getValidClassNames(className)
+        .split(' ')
+        .map((className) => {
+            return styles[className] || className;
+        })
+        .join(' ');
```
anton-otroshchenko: What is the point of it?

frontend/src/bundles/common/components/checkbox/styles.module.scss
```diff
@@ -0,0 +1,9 @@
+.primary {
+    svg {
+        fill: #d5dce8;
```
anton-otroshchenko: Create a variable for this

frontend/src/bundles/common/components/checkbox/styles.module.scss
```diff
@@ -0,0 +1,9 @@
...
+    }
+
+    input:checked + svg {
+        fill: #274f8d;
```
anton-otroshchenko: Same here

anton-otroshchenko (changes requested):

frontend/src/bundles/common/components/checkbox/checkbox.tsx
ptarho: The logic behind it is if class exist in our component style.module.scss it will apply that class from it, otherwise it will just apply pure (global) className

nikita-remeslov:
Conflicts

anton-otroshchenko (approved):

frontend/src/bundles/common/components/checkbox/styles.module.scss
```diff
@@ -0,0 +1,31 @@
+@import url("../../../../assets/css/styles.scss");
```
nikita-remeslov:
```suggestion
```

frontend/src/assets/css/variables.css
```diff
@@ -1,5 +1,9 @@
 :root {
-  --bt-blue-gray: #90a3bf;
-  --bt-input: #d5dce8;
-  --hover: #274F8D;
+    --bt-blue-gray: #90a3bf;
```
nikita-remeslov: my fail, this file should be .scss and imported to styles.scss

nikita-remeslov (changes requested):

frontend/src/assets/css/variables.css
ptarho: So i just need to change extension to scss and leave my variables as it is?

nikita-remeslov (approved):

PR #62, opened 2023-08-22
bt-27: + Create Slider component

### LINKED ISSUE #27: Create Slider component
Image

https://mui.com/material-ui/react-slider/

JemiJe:
Screenshot (569)

I had the same problem described this and took solution in first answer [here](https://github.com/mui/material-ui/issues/20063) 
added styled(Slider)([style options object])


frontend/src/bundles/common/components/slider/slider.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+type Mark = {
+    label?: string;
+    value: number;
+};
+
+type Properties = {
+    label?: string;
+    value?: number;
+    marks?: Mark[] | boolean | undefined;
```
MaksymGavatuk: Why marks is optional and can be boolean ?

MaksymGavatuk: Also you don't need to provide type for undefined like this if you use ? operator
```suggestion
    value?: number;
    marks?: Mark[] | boolean;
```

frontend/src/bundles/common/components/slider/slider.tsx
```diff
@@ -0,0 +1,60 @@
...
+    marks = defaultExperienceMarks,
+    className = classes.sliderContainerStyle,
+}) => {
+    const StyledSlider = styled(Slider)(defaultSliderStyle);
+
+    const getValueLabel = useCallback(
+        (value: number): string | undefined => {
+            return Array.isArray(marks)
+                ? (marks.find((mark) => mark.value === value)?.label as string)
+                : 'undefined';
```
MaksymGavatuk: First of all your function now returns 'undefined' - string instead of undefined, if you want to return undefined.
Also returning undefined is bad practice, you can return null instead of undefined

```suggestion
        (value: number): string | null => {
            return Array.isArray(marks)
                ? (marks.find((mark) => mark.value === value)?.label as string)
                : null;
```

frontend/src/bundles/common/components/slider/slider.tsx
```diff
@@ -0,0 +1,60 @@
...
+                classes.sliderContainerStyle,
+            )}
+        >
+            <span>{label}</span>
+            <StyledSlider
+                aria-label={label}
+                defaultValue={value}
+                marks={marks}
+                step={null}
+                valueLabelDisplay="on"
```
MaksymGavatuk: We don't need to pass _step, valueLabelDisplay_ through props ?

MaksymGavatuk (changes requested):

frontend/src/bundles/common/components/slider/slider.tsx
JemiJe: 1. I thought if we don't provide marks so default values will be used (which already include needed values). that makes component work out of the box without any props provided. but if that prop must be required, I will change it.
2. boolean - it was standard value types for `marks` prop of mui component (described in API Slider), I just put it to avoid possible type errors, but I can remove it, especially in our case we require only `Marks[]`

frontend/src/bundles/common/components/slider/slider.tsx
JemiJe: agree, I will make appropriate changes

frontend/src/bundles/common/components/slider/slider.tsx
JemiJe: 1. `step` I made according to [docs](https://mui.com/material-ui/react-slider/#restricted-values), where stated that `You can restrict the selectable values to those provided with the marks prop with step={null}.`, as I see we don't need to pass step, or  there will be cases when experience could be between values provided by `marks`, for example, we have 1 year and 2 years, but a user can set experience like 1.6 year
2. as I saw in task, we always show label value (1 year, 2 years...), but if that must be optional I can change it. and it would be convenient if you always see what you set in experience

frontend/src/bundles/common/components/slider/slider.tsx
JemiJe: agree, I will remove it

MaksymGavatuk: In my opinion we can leave only Marks[] as required props

JemiJe: ok, I will make it required

frontend/src/bundles/common/components/slider/slider.tsx
MaksymGavatuk: Yes you are absolutely right. In our cases it correct props. But bare in mind, that we create general component which can be used not only for this situation. So you need to pass this variables through props, but make it unrequired and provide default values for this props which you need. That's why we can use this component whatever we want and provide new props for _step, valueLabelDisplay_ which we need in different situations. So main point of doing it is greater flexibility and reuse of the component

JemiJe: ok, I will make changes

MaksymGavatuk:
Did you push all changes ? Can i review ?

JemiJe:
> Did you push all changes ? Can i review ?

yes, if I made it properly (made commit, then push the branch), last commit is [here](https://github.com/BinaryStudioAcademy/bsa-2023-bsa-talents/pull/62/commits/37eabc8f2e58a1c54c13180e47175d823c54954f)

MaksymGavatuk:
> > Did you push all changes ? Can i review ?
> 
> yes, if I made it properly (made commit, then push the branch), last commit is [here](https://github.com/BinaryStudioAcademy/bsa-2023-bsa-talents/pull/62/commits/37eabc8f2e58a1c54c13180e47175d823c54954f)

Okay, thank you. Also if you push changes please re-request review from reviewer because in this case we will get email notification )
image

frontend/src/bundles/common/components/slider/slider.tsx
```diff
@@ -0,0 +1,64 @@
...
+    valueLabelDisplay = 'on',
+}) => {
+    const StyledSlider = styled(Slider)(defaultSliderStyle);
+
+    const getValueLabel = useCallback(
+        (value: number): string | null => {
+            return Array.isArray(marks)
+                ? (marks.find((mark) => mark.value === value)?.label as string)
+                : null;
+        },
```
MaksymGavatuk: Don't use 'as string' casting in this case. Instead find value, check if this item exist and after if not return null in other case return label

frontend/src/bundles/common/components/slider/slider.tsx
```diff
@@ -0,0 +1,64 @@
...
+    value: number;
+};
+
+type Properties = {
+    label?: string;
+    value?: number;
+    className?: string;
+    step?: number | null;
+    valueLabelDisplay?: 'auto' | 'on' | 'off';
+    marks: Mark[];
```
MaksymGavatuk: Change order of props. First of all marks and after that other, because marks is required

frontend/src/bundles/common/components/slider/slider.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+import { getValidClassNames } from '../../helpers/helpers.js';
+import { defaultExperienceMarks } from './default-values.js';
+import classes from './style.module.css';
+import { defaultSliderStyle } from './styles.js';
+
+type Mark = {
+    label?: string;
+    value: number;
+};
```
MaksymGavatuk: Name it more general like Option or something like that, because in this case this name bind us with some context.

frontend/src/bundles/common/components/slider/slider.tsx
```diff
@@ -0,0 +1,64 @@
...
+    className?: string;
+    step?: number | null;
+    valueLabelDisplay?: 'auto' | 'on' | 'off';
+    marks: Mark[];
+};
+
+const CustomSlider: React.FC<Properties> = ({
+    label,
+    value,
+    marks = defaultExperienceMarks,
```
MaksymGavatuk: I think you can delete defaultExperienceMarks or make it like constant for all project if this array used in different part of application, because in this component we create general slider component without any context, so just create constant for variable defaultExperienceMarks, if it needed and move it to general folder, only if we will use it in whole app. After that provide this value as props, but default value for this variable should be empty array.
Also about name, name it please like options or something like that, more general without context.

MaksymGavatuk (changes requested):

frontend/src/bundles/common/components/slider/slider.tsx
JemiJe: I can propose that variant
```javascript
    const getValueLabel = useCallback(
        (value: number): string | null => {
            const label = marks.find((mark) => mark.value === value)?.label;
            return label ?? null;
        },
        [marks],
    );
```

frontend/src/bundles/common/components/slider/slider.tsx
JemiJe: ok, I will just delete it

frontend/src/bundles/common/components/slider/slider.tsx
```diff
@@ -0,0 +1,62 @@
...
+    className = classes.sliderContainerStyle,
+    step = null,
+    valueLabelDisplay = 'on',
+}) => {
+    const StyledSlider = styled(Slider)(defaultSliderStyle);
+
+    const getValueLabel = useCallback(
+        (value: number): string | null => {
+            const label = marks.find((mark) => mark.value === value)?.label;
+            return label ?? null;
```
MaksymGavatuk: Can I propose some changes ? We separate function to logical block, so first we get item from array and next step return value from this item or return null. This is not a critical point, but it is better for the overall logical structure of the code

```suggestion
            const mark = marks.find((mark) => mark.value === value);
            
            return mark?.label ?? null;
```

MaksymGavatuk (changes requested):

frontend/src/bundles/common/components/slider/slider.tsx
MaksymGavatuk: I propose some changes in this variant in other comment

frontend/src/bundles/common/components/slider/slider.tsx
JemiJe: ok, I have tested it, and can add

frontend/src/bundles/common/components/slider/styles.ts
```diff
@@ -0,0 +1,45 @@
...
+        '&:before': {
+            display: 'none',
+        },
+    },
+    '& .MuiSlider-mark': {
+        opacity: 0,
+    },
+};
+
+export { defaultSliderStyle };
```
MaksymGavatuk, line 45: Move it to styles.module.scss and using module styles

MaksymGavatuk (approved):

frontend/src/bundles/common/components/slider/styles.ts
MaksymGavatuk: Move it to styles.module.scss and using module styles

MaksymGavatuk (changes requested):

MaksymGavatuk:
Resolve conflicts also please

frontend/src/bundles/common/components/slider/styles.ts
JemiJe: what is the level of nesting? can I make it like that (2 lvl of nesting):
```scss
    /* parent and inside: */
    &.MuiSlider-markLabel {
        transform: 'translateX(0%)';
        &:nth-of-type(n + 5) {
            display: none;
        }
    }
```

JemiJe: made it, but only via `:global {...}` for nested classes. and eslint didn't pass it because of ':global'

nikita-remeslov: @JemiJe check this comment please - https://github.com/BinaryStudioAcademy/bsa-2023-bsa-talents/pull/77#discussion_r1303459256

frontend/src/bundles/common/components/slider/slider.tsx
```diff
@@ -0,0 +1,62 @@
+import { Box, Slider } from '@mui/material';
```
MaksymGavatuk: Change import like:
```suggestion
import { Box as MUIBox, Slider as MUISlider } from '@mui/material';
```

MaksymGavatuk (changes requested):

frontend/src/bundles/common/components/slider/styles.ts
JemiJe: I used variant proposed by @MaksymGavatuk 
```javascript
// slider.tsx
<Slider
  ...
  classes={{
    markLabel: styles.markLabel
  }}
/>

// styles.module.scss
.markLabel {
    transform: translateX(0%);
}

.markLabel:nth-of-type(n + 5) {
    display: none;
}
```

JemiJe:
1. In unknown cases class root (= MuiSlider-root ) for slider component, when you assign it via classes prop doesn't work. so for slider with (like it was in task) I specified styled directly via style prop, I didn't find way to make it using CSS module.
2. additionally style linted doesn't pass double colon for pseudo-elements, I must put `stylelint-disable selector-pseudo-element-colon-notation`, because of single colon doesn't work.

nikita-remeslov:
@JemiJe check this rule - https://stylelint.io/user-guide/rules/selector-pseudo-element-colon-notation/

frontend/src/bundles/common/components/slider/styles.module.css
nikita-remeslov, line 1: `scss` file

nikita-remeslov: Pay attention to the errors the linter shows you
image

you should fix only `::before`

nikita-remeslov (changes requested):
please merge development branch.

Also, I run your branch locally and don't understand what is wrong. Could you provide more info?

frontend/src/bundles/common/components/slider/styles.module.css
JemiJe: ok, I fixed it

nikita-remeslov:
Conflicts

JemiJe:
> Conflicts

I didn't make it before, but as I understand, I just remove those comments about conflicts, and leave my export of component as it was

nikita-remeslov (approved):

mskmee (approved):

PR #69, opened 2023-08-22
bt-9: create input component

### LINKED ISSUE #9: Create Input component
You should create a component which we will reuse throughout the project

Image

https://mui.com/material-ui/react-text-field/

vladjsp:
input

I tried my best. That was so hard to implement all these variants. Very challenging task. I also will return to it for import colors and typography from shared but for now I think it is done.


frontend/src/bundles/common/components/input/input.tsx
```diff
@@ -1,20 +1,28 @@
...
+
 type Properties<T extends FieldValues> = {
     control: Control<T, null>;
     errors: FieldErrors<T>;
     label: string;
     name: FieldPath<T>;
     placeholder?: string;
-    type?: 'text' | 'email';
+    type?: InputType;
+    disabled?: boolean;
```
anton-otroshchenko: https://github.com/BinaryStudioAcademy/quality-criteria/blob/production/src/javascript.md

B19

anton-otroshchenko (changes requested):

frontend/src/bundles/common/components/input/input.tsx
```diff
@@ -1,20 +1,28 @@
...
     type Control,
     type FieldErrors,
     type FieldPath,
     type FieldValues,
 } from 'react-hook-form';
 
+import searchIcon from '~/assets/img/search.svg';
 import { useFormController } from '~/bundles/common/hooks/hooks.js';
 
+import { getValidClassNames } from '../../helpers/helpers.js';
```
anton-otroshchenko: Use absolute path '~/bundles/common/helpers/helpers.js'

anton-otroshchenko (changes requested):

vladjsp:
I used stylelint-disable in styles one time because that was imposible to overwrite mui helper text color without !important.

vladjsp:
inputs2

vladjsp:
Don't mention top input. It actually has red border while his state == error. Just something happens while I took screenshot.

frontend/src/bundles/common/components/input/styles.module.scss
```diff
@@ -0,0 +1,115 @@
+@import url("../../../../assets/css/styles.scss");
```
lavandosya: I think we don't need this import because we already import styles in common file

vladjsp: Do you mean index.tsx from the frontend root? Okay, I agree. Should I remove it right now or when I will come back to it next time when Josh or Vova fix all variables with propper names?

lavandosya: Maybe it's better to do now but I'm not sure

lavandosya (approved):

frontend/src/bundles/common/components/input/styles.module.scss
nikita-remeslov: It is already imported here - frontend/src/index.tsx

frontend/src/bundles/common/components/input/input.tsx
```diff
@@ -24,21 +32,75 @@ const Input = <T extends FieldValues>({
...
+    );
+    const helperTextStyles = getValidClassNames(
+        styles.helperText,
+        hasError && styles.hasError,
+    );
+
     return (
-        <label>
-            <span>{label}</span>
+        <label className={styles.wrapper}>
```
nikita-remeslov: we don't need label, we just need input component

frontend/src/bundles/common/components/input/styles.module.scss
```diff
@@ -0,0 +1,115 @@
...
+    padding: 11px 14px 11px 10px;
+}
+
+.inputDisabled {
+    pointer-events: none;
+    background-color: var(--light-background-color);
+}
+
+.hasError {
+    /* stylelint-disable */
```
nikita-remeslov: ?

nikita-remeslov (changes requested):

frontend/src/bundles/common/components/input/input.tsx
vladjsp: So should I delete from input component all labels and return only TextField. Like this:
```
return (
            <TextField
                {...field}
                type={type}
                placeholder={placeholder}
                error={hasError}
                helperText={(error as string) || ' '}
                className={textFieldRootStyles}
                InputProps={{
                    className: muiInputStyles,
                    disabled: isDisabled,
                    startAdornment: adornment,
                }}
                inputProps={{ className: htmlInputStyles }}
                FormHelperTextProps={{ className: helperTextStyles }}
            />
    );
```
and remobe label from Props of component?

anton-otroshchenko (approved):

lavandosya (approved):

nikita-remeslov (approved):

PR #70, opened 2023-08-22
bt-66: Add protected route component

### LINKED ISSUE #66: Create Protected route component
Redirect not signed in users from private routes to Sign In page

MChornomaz:


frontend/src/bundles/auth/components/protected-route/protected-route.tsx
```diff
@@ -0,0 +1,23 @@
...
+    }));
+
+    const isAuth = dataStatus === DataStatus.FULFILLED;
+
+    if (!isAuth) {
+        return <Navigate to={AppRoute.SIGN_IN} replace />;
+    }
+
+    return <Route {...props} />;
+};
```
anton-otroshchenko: We will wrap our pages in this component so you also need to pass children

anton-otroshchenko (changes requested):

anton-otroshchenko (approved):

nikita-remeslov (approved):

frontend/src/bundles/auth/components/protected-route/protected-route.tsx
```diff
@@ -0,0 +1,31 @@
...
+import { AppRoute } from '~/bundles/common/enums/app-route.enum.js';
+import { DataStatus } from '~/bundles/common/enums/data-status.enum.js';
+import { useAppSelector } from '~/bundles/common/hooks/hooks.js';
+
+type ProtectedRouteProperties = {
+    children: ReactNode;
+};
+
+const ProtectedRoute: FC<ProtectedRouteProperties> = ({ children }) => {
+    const { dataStatus } = useAppSelector(({ users }) => ({
```
nikita-remeslov: looks like we should check the user. Data can fulfilled without the user
`const hasUser = Boolean(user)`

nikita-remeslov (changes requested):

frontend/src/bundles/auth/components/protected-route/protected-route.tsx
nikita-remeslov:
```js
 //TODO: update when user data will be stored in the store
 const hasUser = false
```

MChornomaz: @nikita-remeslov Do I understand right that for now I should replace the check with const hasUser = false to make it constantly false, until we can get the user?   I made the check for dataStatus for now because I had seen that there is no user in store for now, and I thought that this check would be better, but I can change it to constant "false"

nikita-remeslov: @MChornomaz Yes that's right. Then, we will check the presence of the user in the store, if it is not there, we will redirect to the login page

MChornomaz: @nikita-remeslov  Ok. Thanks, I will change that right after meeting)

frontend/src/bundles/auth/components/protected-route/protected-route.tsx
```diff
@@ -0,0 +1,22 @@
+import { type FC, type ReactNode } from 'react';
+import { Navigate } from 'react-router-dom';
+
+import { AppRoute } from '~/bundles/common/enums/app-route.enum.js';
+
+type ProtectedRouteProperties = {
+    children: ReactNode;
+};
```
MaksymGavatuk: We name this type Properties in other component, please rename it to keep the same naming in whole project
```suggestion
type Properties = {
    children: ReactNode;
};
```

MaksymGavatuk (changes requested):

nikita-remeslov (approved):

PR #77, opened 2023-08-23
bt-6: create chip component

### LINKED ISSUE #6: Create a Chip component which we will use for badges
Image

Image

Image

lavandosya:
I've made common chip component, but got issue with 

`:global(.MuiChip-deleteIcon) {
        width: 13px;
        height: 13px;
        color: var(--text-color-primary);
    }` 

because ESLint does not allow me to use :global. Plain CSS classes do not work for delete-icon and are overwritten by MUI classes.

That's how my chip components look like:

Screenshot 2023-08-22 at 23 26 07


frontend/src/bundles/common/components/chip/chip.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+type Properties = {
+    label: string;
+    onDelete?: () => void;
+};
+
+const Chip = ({ label, onDelete }: Properties): JSX.Element => {
+    return (
+        <MUIChip
+            className={clsx(styles.chip, { [styles['delete-chip']]: onDelete })}
```
MaksymGavatuk: Use please _getValidClassNames_ helpers

```suggestion
            className={getValidClassNames(styles.chip, { [styles['delete-chip']]: onDelete })}
```

frontend/src/bundles/common/components/chip/chip.tsx
```diff
@@ -0,0 +1,23 @@
...
+            label={label}
+            onDelete={onDelete}
+            deleteIcon={onDelete ? <Close /> : undefined}
```
MaksymGavatuk: You can change it to conditional rendering:  
```suggestion
            deleteIcon={onDelete && <Close />}
```

MaksymGavatuk (changes requested):

frontend/src/bundles/common/components/chip/chip.tsx
lavandosya: thanks, I've changed it

MaksymGavatuk:
Also please add component export from general folder components.ts

frontend/src/bundles/common/components/chip/chip.tsx
```diff
@@ -0,0 +1,25 @@
+import { Close } from '@mui/icons-material';
+import { Chip as MUIChip } from '@mui/material';
+
+import { getValidClassNames } from '../../helpers/helpers.js';
```
MaksymGavatuk: Don't use relative path in this case use absolute path like '~/bundles/common/helpers/helpers.js'

MaksymGavatuk (changes requested):

frontend/src/bundles/common/components/chip/chip.tsx
```diff
@@ -0,0 +1,26 @@
...
+    label: string;
+    onDelete?: () => void;
+};
+
+const Chip = ({ label, onDelete }: Properties): JSX.Element => {
+    return (
+        <MUIChip
+            className={getValidClassNames(styles.chip, {
+                [styles['delete-chip']]: onDelete,
+            })}
```
nikita-remeslov: Let's move to the variable. It will be much cleaner

frontend/src/bundles/common/components/chip/chip.tsx
```diff
@@ -0,0 +1,26 @@
...
+            label={label}
+            onDelete={onDelete}
+            deleteIcon={onDelete && <Close />}
+        />
+    );
+};
+
+export { Chip };
```
nikita-remeslov:
```suggestion
import { Close } from '@mui/icons-material';
import { Chip as MUIChip } from '@mui/material';

import { getValidClassNames } from '~/bundles/common/helpers/helpers.js';

import styles from './styles.module.scss';

type Properties = {
    label: string;
    onDelete?: () => void;
};

const Chip = ({ label, onDelete }: Properties): JSX.Element => {
    return (
        <MUIChip
            className={getValidClassNames(styles.chip, {
                [styles['delete-chip']]: onDelete,
            })}
            label={label}
            onDelete={onDelete}
            deleteIcon={onDelete && <Close className={styles.icon} />}
        />
    );
};

export { Chip };

```

frontend/src/bundles/common/components/chip/styles.module.scss
```diff
@@ -0,0 +1,29 @@
...
+}
+
+.chip span {
+    padding: 5px 6px;
+    font-size: var(--font-size-input);
+}
+
+.delete-chip {
+    margin-right: 4px;
+}
```
nikita-remeslov:
```suggestion
.chip {
    gap: 6px;
    width: auto;
    height: 23px;
    margin-right: 9px;
    margin-bottom: 5px;
    color: var(--text-color-primary);
    background-color: var(--tag-main);
    border-radius: 5px;

    & span {
        padding: 5px 6px;
        font-size: var(--font-size-input);
    }

    & .icon {
        width: 13px;
        height: 13px;
        color: var(--text-color-primary);
    }
}

.delete-chip {
    margin-right: 4px;
}
```

nikita-remeslov (changes requested):

frontend/src/bundles/common/components/chip/styles.module.scss
```diff
@@ -0,0 +1,29 @@
+@import url("../../../../assets/css/styles.scss");
+
+/* stylelint-disable selector-pseudo-class-no-unknown */
+
```
nikita-remeslov: Looks like you don't need this code

```suggestion
```

nikita-remeslov (changes requested):

nikita-remeslov (approved):

MaksymGavatuk (approved):

frontend/src/bundles/common/components/chip/chip.tsx
```diff
@@ -0,0 +1,28 @@
...
+import { getValidClassNames } from '~/bundles/common/helpers/helpers.js';
+
+import styles from './styles.module.scss';
+
+type Properties = {
+    label: string;
+    onDelete?: () => void;
+};
+
+const Chip = ({ label, onDelete }: Properties): JSX.Element => {
```
mskmee: We use React.FC in our components, maybe better to make it in one style?

lavandosya: It's a good question, than we need to change chip, input, loader, textarea. Should we do it in bug fix ?

JosuerBague: I don't think this should be a bug fix since it is not a bug, and PR is not merged yet. That is my opinion though, not sure how others feel.

lavandosya: okay, I'll change it

nikita-remeslov: > It's a good question, than we need to change chip, input, loader, textarea. Should we do it in bug fix ?

I got it, yes, we need to make `React.FC`. Could you create a bug ticket?

lavandosya: Yes, of course

PR #79, opened 2023-08-23
bt-21: secure endpoint routing

### LINKED ISSUE #21: Secure Endpoint Routing
Implement private and public endpoints logic

- [x] Add authorization plugin for routes
- [x] Inject signed current user to request (if token is valid)
- [x] Add white (public) routes. For now: sign-in, sign-up only

JosuerBague:
Draft PR,

Waiting for token implementation

Auth Plugin:
image

Auth Plugin Registration:
image


backend/package.json
```diff
@@ -35,6 +35,8 @@
         "convict": "6.2.4",
         "dotenv": "16.3.1",
         "fastify": "4.21.0",
+        "fastify-plugin": "4.5.1",
+        "jose": "4.14.4",
```
MaksymGavatuk, line 38: If you don't use this package remove please this dependency. PR should contain only changes connected with current task. This package should be added in task#20

backend/src/common/server-application/plugins/authorization/authorization.plugin.ts
```diff
@@ -0,0 +1,32 @@
...
+
+type AuthOptions = {
+    routesWhiteList: string[];
+};
+
+const authorizationPlugin: FastifyPluginCallback<AuthOptions> = (
+    fastify,
+    { routesWhiteList },
+) => {
+    fastify.addHook('onRequest', (request, reply) => {
```
MaksymGavatuk: Create special enums like ControllerHook which has ON_REQUEST = 'onRequest' and use this enums in this hook

MaksymGavatuk (changes requested):

MaksymGavatuk:
Also create next folder path src/common/plugins/authorization/authorization.plugin.ts
Move it from server-application

JosuerBague:
Merged changes from bt-20: JWT Auth to this branch to give access to token verification helper.

MaksymGavatuk:
When we approved and merged bt-20, we will check this PR, to avoid checking the same thing twice

backend/src/common/controller/libs/enums/controller-hook.enums.ts
nikita-remeslov, line 1: just create an enum folder in the cintroller. 
Also, create root enum.ts file

backend/src/common/plugins/authorization/authorization.plugin.ts
```diff
@@ -0,0 +1,68 @@
...
+    };
+    routesWhiteList: string[];
+};
+
+declare module 'fastify' {
+    // eslint-disable-next-line @typescript-eslint/consistent-type-definitions
+    interface FastifyRequest {
+        user: UserFindResponseDto;
+    }
+}
```
nikita-remeslov: move it to `src/index.d.ts`

backend/src/common/plugins/authorization/authorization.plugin.ts
```diff
@@ -0,0 +1,68 @@
...
+
+const authorizationPlugin: FastifyPluginCallback<AuthOptions> = (
+    fastify: FastifyInstance,
+    { services, routesWhiteList }: AuthOptions,
+    done,
+) => {
+    fastify.decorateRequest('user', null);
+
+    fastify.addHook(ControllerHooks.ON_REQUEST, async (request, reply) => {
+        try {
```
nikita-remeslov: we don't need `try/catch` block here
just throw an error

backend/src/common/server-application/libs/constants/api.constats.ts
nikita-remeslov, line 1: just contstants folder without libs
also create root file

backend/src/common/plugins/authorization/authorization.plugin.ts
```diff
@@ -0,0 +1,68 @@
...
+            const isWhiteRoute = routesWhiteList.includes(request.routerPath);
```
nikita-remeslov: we should check the method
also, we can move it to server-application helper

backend/src/common/plugins/authorization/authorization.plugin.ts
```diff
@@ -0,0 +1,68 @@
...
+
+            if (isWhiteRoute) {
+                return;
+            }
+
+            if (request.headers.authorization) {
```
nikita-remeslov: Get token, and if we do not have the token throw an error

backend/src/common/plugins/authorization/authorization.plugin.ts
```diff
@@ -0,0 +1,68 @@
...
+                const { payload } = await tokenService.decode(
```
nikita-remeslov: you can get tokenService for services (the same as you did with userService)

backend/src/common/plugins/authorization/authorization.plugin.ts
```diff
@@ -0,0 +1,68 @@
...
+                    request.headers.authorization,
+                );
+                const { userService } = services;
+
+                const authorizedUser = await userService.findById(
+                    payload.id as number,
+                );
+                if (!authorizedUser) {
+                    throw new Error(
+                        'You are not authorized to access this route.',
```
nikita-remeslov: for errors, we should create an enum. we need it for frontend, mobile

backend/src/common/server-application/server-app-base.ts
```diff
@@ -89,6 +92,13 @@ class ServerAppBase implements ServerApp {
                 await this.app.register(swaggerUi, {
                     routePrefix: `${it.version}/documentation`,
                 });
+
+                await this.app.register(authorization, {
```
nikita-remeslov: this is plugin not middleware

nikita-remeslov (changes requested):

backend/src/common/plugins/authorization/authorization.plugin.ts
JosuerBague: I tried moving this to `src/index.d.ts` but TS can't seem to pick up on it.

backend/src/common/plugins/authorization/authorization.plugin.ts
JosuerBague: Hello, I am not sure what this means. Isn't it already handled in the current code?

```
if (request.headers.authorization) {
        // code here
} else {
    throw new Error(AuthorizationErrorMessages.NOT_AUTHORIZED);
}

```

backend/src/common/plugins/authorization/authorization.plugin.ts
JosuerBague: Just to clarify,
You want to make isWhiteRoute a helper and it should also check the request method used?

If so, should I only check for GET methods and only allow such methods?

backend/src/common/plugins/authorization/authorization.plugin.ts
JosuerBague: Current iteration:
image

backend/src/common/plugins/authorization/authorization.plugin.ts
nikita-remeslov: yes. you can add something like `checkWhiteRoute`

All routes should have path and method

backend/src/common/plugins/authorization/authorization.plugin.ts
nikita-remeslov: I don't like `if` block in the `if` block.

```js
    const [, token] = request.headers.authorization?.split(' ') ?? [];

    if (!token) {
        throw error;
    }

    const authorizedUser;

    if (!authorizedUser) {
        throw error;
    }
```

backend/src/common/plugins/authorization/authorization.plugin.ts
JosuerBague: checkWhiteRoute helper:

image

backend/src/common/plugins/authorization/authorization.plugin.ts
JosuerBague: New Iteration:

image

anton-otroshchenko (approved):

backend/src/index.d.ts
nikita-remeslov, line 1: add to backend eslintrc.yml
```
  - files:
    - src/index.d.ts
    rules:
      '@typescript-eslint/consistent-type-definitions':
        - off
```

backend/src/index.d.ts
```diff
@@ -0,0 +1,8 @@
+import { type UserFindResponseDto } from './bundles/users/types/types.js';
+
+declare module 'fastify' {
+    // eslint-disable-next-line @typescript-eslint/consistent-type-definitions
```
nikita-remeslov:
```suggestion
```

shared/src/enums/authorization-error-messages.enum.ts
nikita-remeslov, line 1: error-messages.enum.ts

backend/src/common/plugins/authorization/authorization.plugin.ts
```diff
@@ -0,0 +1,58 @@
...
+        }
+
+        const { userService, tokenService } = services;
+
+        const { payload } = await tokenService.decode(authorization as string);
+
+        const authorizedUser = await userService.findById(payload.id as number);
+        if (!authorizedUser) {
+            throw new Error(AuthorizationErrorMessages.NOT_AUTHORIZED);
+        }
```
nikita-remeslov:
```suggestion
        const { userService, tokenService } = services;
        const { payload } = await tokenService.decode(authorization as string);
        const authorizedUser = await userService.findById(payload.id as number);
        
        if (!authorizedUser) {
            throw new Error(AuthorizationErrorMessages.NOT_AUTHORIZED);
        }
```

nikita-remeslov: Not a blocker

nikita-remeslov (changes requested):

backend/src/index.d.ts
```diff
@@ -0,0 +1,8 @@
...
+    interface FastifyRequest {
+        user: UserFindResponseDto;
```
MaksymGavatuk: I think in our case it should be optional field, because in WHITE endpoints req.user = undefined
```suggestion
        user?: UserFindResponseDto;
```

MaksymGavatuk (changes requested):

nikita-remeslov (approved):

nikita-remeslov (approved):

PR #82, opened 2023-08-23
bt-61: create overlay component | development

### LINKED ISSUE #61: Create Overlay component
It should be a component that shows screen data but disables all possible touchable items. It also should have a spinner in the center. Also, this component should be fullscreen and almost transparent so that the user can see the screen content.

Example of usage: 
1. User makes some async action (login for example)
2. While waiting for a response user sees Overlay with a spinner and cannot press buttons

mskmee:
image

Made misstake and create branch without task prefix, can't delete it (


mobile/src/bundles/common/components/overlay/overlay.tsx
```diff
@@ -0,0 +1,37 @@
...
+    color?: string;
+};
+const Overlay: React.FC<OverlayProperties> = ({
+    isActive,
+    color = Color.PRIMARY,
+}) => {
+    if (!isActive) {
+        return null;
+    }
+    return (
```
likeri29, line 23: I think better approach is:
`return ({isActive ? <View
            style={[
                globalStyles.width100,
                globalStyles.height100,
                globalStyles.justifyContentCenter,
                styles.wrapper,
            ]}
        >
            <ActivityIndicator color={color} size="large" />
        </View> : null})`

mskmee: Always write like this, if we will write in syntax like you write, I'll change component

nikita-remeslov: @likeri29 syntax like this more readable, as for me
so I think it's best to stick with it!

@mskmee I am with you 👍

nikita-remeslov (approved):

mobile/src/bundles/common/components/overlay/overlay.tsx
```diff
@@ -0,0 +1,37 @@
...
+        <View
+            style={[
+                globalStyles.width100,
+                globalStyles.height100,
```
katerynakostikova, line 28:
```suggestion
                globalStyles.flex1,
```

mobile/src/bundles/common/components/overlay/overlay.tsx
```diff
@@ -0,0 +1,37 @@
...
+};
+const Overlay: React.FC<OverlayProperties> = ({
+    isActive,
+    color = Color.PRIMARY,
+}) => {
+    if (!isActive) {
+        return null;
+    }
+    return (
+        <View
```
katerynakostikova, line 24:
```suggestion
        <View
        pointerEvents="box-only"
```

mobile/src/bundles/common/components/overlay/overlay.tsx
mskmee: Flex: 1 didnt resize absolute component, or i should use it in another way?

mobile/src/bundles/common/components/overlay/overlay.tsx
mskmee: fix

katerynakostikova (approved):

PR #83, opened 2023-08-23
bt-30: Create Bottom tabs navigation

anogire:
I have made a separate AuthNavigation, which should include all pages for authorized users (maybe there will be more, not just those in the bottom tab).
So, app.tsx will probably have a switch to AuthNavigation or RootNavigation.

PR #84, opened 2023-08-23
bt-31: Create Drawer navigation Screen header

### LINKED ISSUE #31: Create Drawer navigation Screen header
Create a Drawer navigation component (Only the top screen header for now) 

Image

likeri29:
Waiting when icons will be merged in the development branch. The header looks like this: 
image


mobile/src/bundles/common/components/header/header.tsx
```diff
@@ -0,0 +1,41 @@
+import React from 'react';
+
+import { Text, View } from '~/bundles/common/components/components';
+import { TextCategory } from '~/bundles/common/enums/enums';
+import { globalStyles } from '~/bundles/common/styles/global-styles';
+
+import { styles } from './styles';
+
+const Header = (): JSX.Element => {
```
katerynakostikova: This header should be a separate navigation component named OnboardingNavigation (for example) and have a Drawer navigation 
Investigate some variants of how you can customise the screen header in the Drawer navigator. See [this PR](https://github.com/BinaryStudioAcademy/bsa-2023-bsa-talents/pull/83#pullrequestreview-1592867827) comments also to see how the navigation structure should look like

katerynakostikova (changes requested):

mobile/src/bundles/common/components/header/header.tsx
likeri29: While my research, I found out that we can pass another component as a header for Drawer Navigator, like this: 
image
So basically, I can use this created component by passing navigation as a props and by click on the icon drawer opens, like this (I will fix this ts errors): 
image

Is it a good approach?

likeri29:
I'll remove the OnboardingNavigation from the App.tsx

mobile/src/bundles/common/components/header/header.tsx
```diff
@@ -0,0 +1,52 @@
...
+import Icon from 'react-native-vector-icons/MaterialIcons';
+
+import { Pressable, Text, View } from '~/bundles/common/components/components';
+import { TextCategory } from '~/bundles/common/enums/enums';
+import { globalStyles } from '~/bundles/common/styles/global-styles';
+
+import { styles } from './styles';
+
+type Header = {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
likeri29: I will change it

mobile/src/bundles/common/components/header/header.tsx
katerynakostikova: Yes, good

mobile/src/app/app.tsx
```diff
@@ -7,14 +7,16 @@ import { Provider as StoreProvider } from 'react-redux';
...
+import { OnboardingNavigation } from '~/navigations/onboarding-navigation/onboarding-navigation';
 
 const App: FC = () => {
     return (
         <StoreProvider store={store.instance}>
             <GestureHandlerRootView style={globalStyles.flex1}>
                 <NavigationContainer>
-                    <RootNavigation />
+                    {/* <RootNavigation /> */}
+                    <OnboardingNavigation />
```
katerynakostikova: Let's leave root nav here, you can return OnboardingNavigation conditionally in RootNavigation

katerynakostikova: Let's add some flag which for now we will change manually isProfileComplete (for ex.)

mobile/src/bundles/common/components/header/header.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { TextCategory } from '~/bundles/common/enums/enums';
+import { globalStyles } from '~/bundles/common/styles/global-styles';
+
+import { styles } from './styles';
+
+type Header = {
+    navigation: DrawerNavigationProp<ParamListBase>;
+};
+
+const Header = ({ navigation }: Header): JSX.Element => {
```
katerynakostikova:
```suggestion
const Header: React.FC<DrawerNavigationProp> = ({ navigation })=> {
```

mobile/src/bundles/common/components/header/header.tsx
```diff
@@ -0,0 +1,55 @@
...
+                globalStyles.flexDirectionRow,
+                globalStyles.alignItemsCenter,
+                globalStyles.justifyContentCenter,
+            ]}
+        >
+            <Pressable
+                style={styles.icon}
+                onPress={(): void => {
+                    navigation.openDrawer();
+                }}
```
katerynakostikova:
```suggestion
                onPress={navigation.openDrawer}
```

mobile/src/bundles/common/components/header/styles.ts
```diff
@@ -0,0 +1,29 @@
+import { StyleSheet } from 'react-native';
+
+import { Color } from '~/bundles/common/enums/enums';
+
+const styles = StyleSheet.create({
+    header: {
+        backgroundColor: Color.TEXT,
+        height: 100,
```
katerynakostikova:
```suggestion
```
globalStyles

mobile/src/navigations/onboarding-navigation/onboarding-navigation.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+const OnboardingNavigation: React.FC = () => {
+    return (
+        <Drawer.Navigator
+            screenOptions={{
+                headerShown: true,
+                header: ({ navigation }): React.ReactNode => {
+                    return <Header navigation={navigation} />;
+                },
+            }}
```
katerynakostikova:
```suggestion
                header: ({ navigation }): React.ReactNode => (
                    <Header navigation={navigation} />
                ),
            }}
```

mobile/src/navigations/onboarding-navigation/onboarding-navigation.tsx
```diff
@@ -0,0 +1,26 @@
...
+        >
+            <Drawer.Screen name={RootScreenName.SIGN_UP} component={Auth} />
+            <Drawer.Screen name={RootScreenName.SIGN_IN} component={Auth} />
```
katerynakostikova: It should return some other screens (you can create empty screens for now)

mobile/src/app/app.tsx
katerynakostikova: See this PR:  [https://github.com/BinaryStudioAcademy/bsa-2023-bsa-talents/pull/83](https://github.com/BinaryStudioAcademy/bsa-2023-bsa-talents/pull/83)

mobile/src/bundles/common/components/header/header.tsx
likeri29: Is it ok if I use openDrawer like this, because when I use your suggested variant I can't commit changes (such error: @typescript-eslint/unbound-method)?
image

mobile/src/bundles/common/components/header/styles.ts
likeri29: I need to create this style? Because we do not have it there

mobile/src/app/app.tsx
likeri29: I am trying to do like this, but when I save prettier delets parenthesis and then liner is complaining. Do you have any suggestions about this?

image

mobile/src/bundles/common/components/header/styles.ts
katerynakostikova: We have height 100 style 
`style={[styles.header, globalStyles.h100]}`

mobile/src/bundles/common/components/header/header.tsx
katerynakostikova: Yes, it's ok

mobile/src/bundles/common/components/header/styles.ts
likeri29: We have only height100 which is 100%
image

mobile/src/app/app.tsx
katerynakostikova: IsProfileComplete should be inside signed in user, firstly sign in check, then profile complete

mobile/src/bundles/common/components/header/styles.ts
katerynakostikova: Oh, sorry, I thought we have % there, then leave it as it

mobile/src/app/app.tsx
katerynakostikova: Add `// prettier-ignore` before return

likeri29: So, leave as I did and just add this `// prettier-ignore`?

likeri29:
Drawer header looks like this
header
drawer

mobile/src/bundles/users/screens/step-one.tsx
anogire, line 1: Maybe we should call these screens "profile", "bsa-badges", ... instead of "step-one", ... ?
Because, as I understand from task #104, I'm doing just this "profile" screen.

likeri29: Yes, it's temporarily names

likeri29: I will rename these two screens now

likeri29: Renamed

mobile/src/bundles/common/components/components.tsx
```diff
@@ -1,5 +1,6 @@
 export { Button } from './button/button';
 export { FormField } from './form-field/form-field';
+export { Header } from './header/header';
```
katerynakostikova: Let's move header to the components folder, in the onboarding nav folder

mobile/src/bundles/common/components/header/header.tsx
```diff
@@ -0,0 +1,47 @@
+import { type DrawerNavigationProp } from '@react-navigation/drawer';
+import { type ParamListBase } from '@react-navigation/native';
+import React from 'react';
```
katerynakostikova:
```suggestion
import React from 'react';
import { type DrawerNavigationProp } from '@react-navigation/drawer';
import { type ParamListBase } from '@react-navigation/native';
```

mobile/src/bundles/common/components/header/header.tsx
```diff
@@ -0,0 +1,47 @@
...
+                styles.header,
+                globalStyles.flexDirectionRow,
+                globalStyles.alignItemsCenter,
+                globalStyles.justifyContentCenter,
+            ]}
+        >
+            <Pressable style={styles.icon} onPress={openDrawer}>
+                <Icon name="menu" size={30} color="#fff" />
+            </Pressable>
+            <Text category={TextCategory.H5} style={{ color: '#fff' }}>
```
katerynakostikova:
```suggestion
            <Text category={TextCategory.H5} style={styles.title}>
```

mobile/src/bundles/common/enums/navigation/drawer-screen-name.enum.ts
```diff
@@ -0,0 +1,6 @@
+const DrawerScreenName = {
```
katerynakostikova:
```suggestion
const OnboardingScreenName = {
```

mobile/src/bundles/common/enums/navigation/root-screen-name.enum.ts
```diff
@@ -1,6 +1,7 @@
 const RootScreenName = {
     AUTH_ROOT_ROUTE: 'Auth Root Route',
     MAIN_ROOT_ROUTE: 'Main Root Route',
+    DRAWER_ROOT_ROUTE: 'Drawer Root Route',
```
katerynakostikova: ONBOARDING_ROOT_ROUTE

mobile/src/bundles/common/types/navigation/navigation.ts
```diff
@@ -1,4 +1,5 @@
 export { type AuthNavigationParameterList } from './auth-navigation-parameter-list.type';
 export { type BottomTabNavigationParameterList } from './bottom-tab-navigation-parameter-list.type';
+export { type DrawerNavigationParameterList } from './drawer-navigation-parameter-list-type';
```
katerynakostikova:
```suggestion
export { type OnboardingNavigationParameterList } from './drawer-navigation-parameter-list-type';
```

mobile/src/bundles/common/components/header/header.tsx
likeri29: I've already tried to do this, but linter is complaining if to move these types below

illia-dobro (commented):
Looks good on screen!

But I have a question: Do we take styles from Figma, such as sizes? I see different sizes here. For example, the circle ('steps') on Figma is 56px, but here it is 75. Do we use some ratio or something like that?

mobile/src/bundles/common/components/header/header.tsx
likeri29: The same situation we have in the root navigation

likeri29:
I can change it 65, which is look better, but 56 will be too small
image

katerynakostikova (approved):

anogire (approved):

illia-dobro (approved):

PR #95, opened 2023-08-24
bt-92: * apply className to FormControlLabel

### LINKED ISSUE #92: Fix checkbox label customization
**Description:**
Right now when the checkbox component created with a label className applied to mui checkbox directly and there is no way to customize label properties.  To fix that we need to apply className on FormControlLabel when checkbox is created with label.

**Preconditions:**
None
**Environment:**
Browser

**Steps to reproduce:**

1.Pass className that is sopposed to rewrite label styles to Checkbox component

**Actual Result:**
Impossible to customize label styles

**Expected Result:**
Label can be customized with className property

**Visual Proof (screenshots, videos, text):**
Label is not customized

**Severity:**
Blocker (S1)

ptarho:
I just moved to FormControlLabel if checkbox is created with label. Because before it was impossible to customize label style.
Now we can do it by changing MUI classname  in next way:
Screenshot_8
Screenshot_4
Screenshot_3

### DIFF
--- a/frontend/src/bundles/common/components/checkbox/checkbox.tsx
+++ b/frontend/src/bundles/common/components/checkbox/checkbox.tsx
@@ -47,11 +47,11 @@ const Checkbox: React.FC<Properties> = ({
                     checked={isChecked}
                     required={isRequired}
                     disabled={isDisabled}
-                    className={getValidClassNames(className)}
                     icon={checkboxIcon}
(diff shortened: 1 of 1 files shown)


Simpsonkk:
It works well

Simpsonkk (approved):

nikita-remeslov (approved):

PR #99, opened 2023-08-24
bt-51: + Create error handling

### LINKED ISSUE #51: Create error handling
Add store middleware
Create notification service using https://www.npmjs.com/package/react-toastify

JemiJe:
1. I change `middleware:` property value of `Store` `constructor` as it was in mini-project, in that case my middleware can catch actions, it doesn't work if I put it in `extraArguments`
2. instead of arrow funcs currying, linter forced me to use regular functions because of
[that rule](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v48.0.1/docs/rules/consistent-function-scoping.md)
it could be `const errorService: Middleware = () => (next) => (action: AnyAction) => {...}`
3. to avoid dublication of notifications I used `{ toastId: message }`

Screenshot (574)


anton-otroshchenko (approved):
Looks good, but please resolve conflicts

JemiJe:
> Looks good, but please resolve conflicts

resolved

frontend/src/framework/store/store.package.ts
```diff
@@ -40,13 +42,14 @@ class Store {
                 auth: authReducer,
                 users: usersReducer,
             },
-            middleware: (getDefaultMiddleware) => {
-                return getDefaultMiddleware({
+            middleware: (getDefaultMiddleware) => [
+                ...getDefaultMiddleware({
                     thunk: {
                         extraArgument: this.extraArguments,
```
nikita-remeslov, line 56: we should add it to extra, in this case, we can call notification in the store

frontend/src/framework/store/middlewares/error-handler/error-handler.middleware.ts
```diff
@@ -0,0 +1,19 @@
...
+import { isRejected } from '@reduxjs/toolkit';
+
+import { notificationService } from '~/services/services.js';
+
+const errorHandler: Middleware = function () {
+    return function (next) {
+        return function (action: AnyAction) {
+            if (isRejected(action)) {
+                const message = action.error.message;
+                notificationService.error(message);
```
nikita-remeslov: you dont need to call the service directly. Create action in the app/store and call action

frontend/src/services/notification/notification.service.ts
```diff
@@ -0,0 +1,22 @@
...
+    }
+    public success(message: Message): void {
+        toast.success(message, { toastId: message });
+    }
+    public warn(message: Message): void {
+        toast.warn(message, { toastId: message });
+    }
+    public error(message: Message): void {
+        toast.error(message, { toastId: message });
+    }
```
nikita-remeslov: create private method show and use it in the public methods

```js
  private show(message: string, parameters: NotificatinOptions): void {
    toast(message, parameters);
  }
```

also create enum for notification type

frontend/src/services/notification/notification.service.ts
nikita-remeslov, line 1: better to make it as a package, to be consistency 
notification.package.ts

nikita-remeslov (changes requested):

frontend/src/framework/store/store.package.ts
JemiJe: as I understand, additionally add service to extra, but leave middleware and dispatch error from it?

frontend/src/framework/store/middlewares/error-handler/error-handler.middleware.ts
JemiJe: create another action and call it inside middleware in the way of `dispatch(notificationActions.error())`?

JemiJe: error prevent me call dispatch from middleware

nikita-remeslov:
Conflicts

frontend/src/app/store/slice.ts
```diff
@@ -0,0 +1,33 @@
...
+        builder.addCase(notify.pending, (state) => {
+            state.dataStatus = DataStatus.PENDING;
+        });
+        builder.addCase(notify.fulfilled, (state) => {
+            state.dataStatus = DataStatus.FULFILLED;
+        });
+        builder.addCase(notify.rejected, (state) => {
+            state.dataStatus = DataStatus.REJECTED;
+        });
+    },
```
nikita-remeslov: I think we don't need extraReducers

nikita-remeslov: and also fix CI errors

nikita-remeslov (changes requested):

nikita-remeslov (approved):

PR #88, opened 2023-08-24
bt-20: jwt auth

JosuerBague:
Tasks done:

- JWT creation using jose
- attaching token to sign-up route
- storing token into local storage
-  creation of helper verify-token

Not done:
- attaching token to sign-in route

image

Issue #86, opened 2023-08-23
Sign-up


MaksymGavatuk:
Finilize please sign-up process

- [x] check and implement all TODO in sign-up process (password hash, ... etc)
- [x] check sign-up flow

Monyster:
What should be used to encrypt the password?
I think it is bcrypt, am I right?

Where can I store a constant value like SALT_ROUNDS?

I also asked in Slack: https://bsa2023.slack.com/archives/C05NY1E7CBS/p1692820941630339

Monyster:
I complete passwordSalt, passwordHash.

anna9314:
Tested
Got Status Code: 201


![Image](https://github.com/BinaryStudioAcademy/bsa-2023-bsa-talents/assets/63346295/1062be35-9600-4885-b004-2323e1d78be9)

Got Status Code: 400


![Image](https://github.com/BinaryStudioAcademy/bsa-2023-bsa-talents/assets/63346295/c7a43e0b-f801-449f-b417-4df8f20e77e2)

PR #100, opened 2023-08-24
bt-19: + implement sign in on server

ptarho:
This is just a first general implementation, because jwt token and verifyLoginCredentials should be done when I will get code from other tasks. Also I have a bunch of questions like where to throw an errors, do I need to replace error messages with corresponding enums or we can leave it for now, ect
  
email_validation
email_wrong
correct
password_wrong


backend/src/bundles/auth/auth.controller.ts
```diff
@@ -76,6 +93,51 @@ class AuthController extends ControllerBase {
             payload: await this.authService.signUp(options.body),
         };
     }
+    /**
+     * @swagger
+     * /auth/sign-in:
+     *    post:
```
MaksymGavatuk, line 101: Did you check swagger api for endpoint ? That is, did you try to make request using swagger ?

shared/src/bundles/users/types/user-sign-in-response-dto.type.ts
```diff
@@ -0,0 +1,7 @@
+type UserSignInResponseDto = {
+    id: number;
+    email: string;
+    token: string;
+};
```
MaksymGavatuk: Leave please only token as I write in comment for jwt auth PR

backend/src/bundles/auth/auth.service.ts
```diff
@@ -1,21 +1,60 @@
...
+        const user = await this.userRepository.findByEmail(email);
+
+        if (!user) {
+            throw new Error('Incorrect email');
+        }
+
+        // const isEqualPassword = true await cryptCompare(password, password);
+        // if (!isEqualPassword) {
+        //     throw new Error('Password not match');
+        // }
```
MaksymGavatuk: This is TODO flow, or you forgot to delete this lines ?

MaksymGavatuk (changes requested):

backend/src/bundles/auth/auth.service.ts
Monyster: I think it's a TODO because I'm implement cryptCompare in my task #86

backend/src/bundles/auth/auth.controller.ts
```diff
@@ -33,6 +36,20 @@ class AuthController extends ControllerBase {
                     }>,
                 ),
         });
+
+        this.addRoute({
+            path: AuthApiPath.SIGN_IN,
+            method: 'POST',
+            validation: {
+                body: userSignUpValidationSchema,
```
JosuerBague: perhaps this should be userSignInValidationSchema?

backend/src/bundles/auth/auth.controller.ts
```diff
@@ -33,6 +36,20 @@ class AuthController extends ControllerBase {
...
+            },
+            handler: (options) =>
+                this.signIn(
+                    options as ApiHandlerOptions<{
+                        body: UserSignUpRequestDto;
```
JosuerBague: might be UserSignInResponseDto

backend/src/bundles/auth/auth.ts
```diff
@@ -1,10 +1,10 @@
-import { userService } from '~/bundles/users/users.js';
+import { userRepository, userService } from '~/bundles/users/users.js';
 import { logger } from '~/common/logger/logger.js';
 
 import { AuthController } from './auth.controller.js';
 import { AuthService } from './auth.service.js';
 
-const authService = new AuthService(userService);
+const authService = new AuthService(userService, userRepository);
```
JosuerBague: I am not sure if we need to give authService direct access to userRepository since useRepository is accessible via userService? I am not sure though.

backend/src/bundles/auth/auth.service.ts
ptarho: I left this code to show that this part is not finished, becuse I need to use cryptCompare or some analogue that was not implemented at that time. Because currently it`s just check if password='HASH', as it was hardcoded inside sign-up route

backend/src/bundles/auth/auth.controller.ts
ptarho: Yeah request are working in swagger ui

backend/src/bundles/auth/auth.ts
anton-otroshchenko: > I am not sure if we need to give authService direct access to userRepository since useRepository is accessible via userService? I am not sure though.

I think you are right. You can create a method in userService which accesses userRepository since you alredy using userService in this class

ptarho: The main reason I imported repository is because I wanted to make findByEmail method inside user-service return just user object if we will need to use it, while inside auth-service we need userEntity with password hash

ptarho: Maybe in this case we can delete user-service and use only user-repository?

backend/src/bundles/users/user.service.ts
```diff
@@ -19,6 +20,30 @@ class UserService implements Service {
         return Promise.resolve(null);
     }
 
+    public async findByEmail(
+        email: UserEntity['email'],
+    ): Promise<UserGetOneItemResponseDto> {
+        const user = await this.userRepository.findByEmail(email);
```
Monyster: I guess this is similar to https://github.com/BinaryStudioAcademy/bsa-2023-bsa-talents/pull/88#discussion_r1305874025
```suggestion
        const user = await this.userRepository.find({email});
```

backend/src/bundles/users/user.service.ts
```diff
@@ -19,6 +20,30 @@ class UserService implements Service {
...
+
+        if (!user) {
+            throw new Error('Incorrect email');
```
Monyster: Change new Error to any exception or create new in folder /shared/src/framework/exceptions

backend/src/bundles/users/user.service.ts
```diff
@@ -19,6 +20,30 @@ class UserService implements Service {
...
+        }
+
+        return user.toObject();
+    }
+
+    public async findById(
+        id: UserEntity['id'],
+    ): Promise<UserGetOneItemResponseDto> {
+        const user = await this.userRepository.findById(id);
```
Monyster: I guess this is similar to https://github.com/BinaryStudioAcademy/bsa-2023-bsa-talents/pull/88#discussion_r1305874025
```suggestion
        const user = await this.userRepository.find({id});
```

backend/src/bundles/auth/auth.service.ts
```diff
@@ -1,21 +1,60 @@
...
+        if (password !== 'HASH') {
+            throw new Error('Password not match');
+        }
+
+        return user.toObject();
+    }
+
+    public async login({ id }: { id: number }): Promise<UserSignInResponseDto> {
```
Monyster: Everywhere you use signIn, but here you use the login function
```suggestion
    public async signIn({ id }: { id: number }): Promise<UserSignInResponseDto> {
```

backend/src/bundles/auth/auth.controller.ts
```diff
@@ -76,6 +93,51 @@ class AuthController extends ControllerBase {
...
+     */
+    private async signIn(
+        options: ApiHandlerOptions<{
+            body: UserSignInRequestDto;
+        }>,
+    ): Promise<ApiHandlerResponse> {
+        const user = await this.authService.verifyLoginCredentials(
+            options.body,
+        );
+        const login = await this.authService.login(user);
```
Monyster:
```suggestion
        const signIn= await this.authService.signIn(user);
```

backend/src/bundles/auth/auth.controller.ts
```diff
@@ -76,6 +93,51 @@ class AuthController extends ControllerBase {
...
+        return {
+            status: HttpCode.OK,
+            payload: login,
```
Monyster:
```suggestion
            payload: signIn,
```

backend/src/bundles/auth/auth.service.ts
Monyster: You can try to merge my branch #86, but my branch also uses JWT and this branch too, so it may take a long time to compare and fix.

cryptCompare already included in /backend/src/common/helpers

Monyster (changes requested):

backend/src/bundles/auth/auth.ts
JosuerBague: > The main reason I imported repository is because I wanted to make findByEmail method inside user-service return just user object if we will need to use it, while inside auth-service we need userEntity with password hash

We can still achieve this with just userService exposed to authService.

You can return a UserEntity via useRepo and then return user Object by running UserEntity.toObject and returning result from service.

See [userService.findByEmail](https://github.com/BinaryStudioAcademy/bsa-2023-bsa-talents/pull/88/files) sample flow.

nikita-remeslov:
@ptarho merge development branch. We merged JWT to development

backend/src/bundles/users/user.entity.ts
```diff
@@ -65,10 +65,12 @@ class UserEntity implements Entity {
     public toObject(): {
         id: number;
         email: string;
+        passwordHash: string;
```
Monyster, line 57: Remove passwordHash.
Depending on the return statement https://github.com/BinaryStudioAcademy/bsa-2023-bsa-talents/blob/d7721ebfa699a87a1ac4d8578dc1c5053b991efe/backend/src/bundles/users/user.service.ts#L42C14-L42C14, we can return passwords to the client in the browser. I think this is a bad idea.

ptarho: I use findByEmail method inside auth-service to validate password. 
If I will remove it from return value I coludn`t use it.
We can switch usage of user-service into user-repository inside auth-service to fix that, but I am not sure about that

Monyster: Oh, to compare the password, we need return it. Sorry, my mistake.

shared/tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
     "extends": "../tsconfig.json",
-    "include": ["src"],
+    "include": ["src", "../backend/src/common/helpers/generate-salt"],
```
JosuerBague: I think this might be causing issues?
Because when I remove this from tsconfig and run 'npm run build' inside shared it correctly builds the shared folder.

I might be wrong though.

backend/src/common/helpers/crypt-compare/crypt-compare.helper.ts
```diff
@@ -0,0 +1,10 @@
+import { compare } from 'bcrypt';
+
+const cryptCompare = async (
+    data: string,
+    encrypted: string,
+): Promise<boolean> => {
+    return await compare(data, encrypted);
```
MaksymGavatuk: You don't need to _return await_ 
I have good article about it http://hassansin.github.io/Why-return-await-Is-a-Bad-Idea

backend/src/common/helpers/encrypt/encrypt.helper.ts
```diff
@@ -0,0 +1,7 @@
+import { hash } from 'bcrypt';
+
+const encrypt = async (data: string, rounds: number): Promise<string> => {
+    return await hash(data, rounds);
```
MaksymGavatuk: The same http://hassansin.github.io/Why-return-await-Is-a-Bad-Idea

backend/src/common/helpers/generate-salt/generate-salt.helper.ts
```diff
@@ -0,0 +1,7 @@
+import { genSalt } from 'bcrypt';
+
+const generateSalt = async (rounds: number): Promise<string> => {
+    return await genSalt(rounds);
```
MaksymGavatuk: The same http://hassansin.github.io/Why-return-await-Is-a-Bad-Idea

backend/src/bundles/users/user.service.ts
```diff
@@ -43,14 +45,25 @@ class UserService implements Service {
         };
     }
 
-    public async create(payload: UserSignUpRequestDto): Promise<UserEntity> {
-        return this.userRepository.create(
+    public async create(
+        payload: UserSignUpRequestDto,
+    ): Promise<UserFindResponseDto> {
+        const { PASSWORD_SALT_ROUNDS } = config.ENV.CRYPT;
```
MaksymGavatuk: save PASSWORD_SALT_ROUNDS as private field for UserService class and pass it in constructor instead of using from import

MaksymGavatuk (changes requested):

backend/src/bundles/users/user.service.ts
ptarho: @MaksymGavatuk done

backend/src/bundles/auth/auth.service.ts
```diff
@@ -1,33 +1,28 @@
...
-        const user = await this.userService.findById(id);
-
+    public async signIn(
+        userRequestDto: UserSignInRequestDto,
+    ): Promise<UserSignInResponseDto> {
+        const user = await this.verifyLoginCredentials(userRequestDto);
         return {
-            ...user,
-            token: await tokenService.create({ id }),
+            token: await tokenService.create({ id: user.id }),
```
nikita-remeslov, line 25: why did you remove the user?

backend/src/bundles/auth/auth.service.ts
```diff
@@ -46,40 +41,36 @@ class AuthService {
         }
 
         const user = await this.userService.create(userRequestDto);
-        const { id } = user;
-
         return {
-            ...user,
-            token: await tokenService.create({ id }),
+            token: await tokenService.create({ id: user.id }),
```
nikita-remeslov, line 45: why did you remove the user?

backend/src/common/helpers/encrypt/encrypt.helper.ts
nikita-remeslov, line 1: you dont need a helper. You already have token service

nikita-remeslov (changes requested):

nikita-remeslov:
@ptarho Pay attention to decode token logic - https://github.com/BinaryStudioAcademy/bsa-2023-bsa-talents/pull/88/files

backend/src/bundles/auth/auth.service.ts
ptarho: https://github.com/BinaryStudioAcademy/bsa-2023-bsa-talents/pull/100#discussion_r1304781021

backend/src/common/helpers/encrypt/encrypt.helper.ts
ptarho: This service is used for jwt token, but helpers is for bcrypt part

PR #105, opened 2023-08-25
bt-87: configure file uploading

### LINKED ISSUE #87: Configure file uploading
Implement file uploading for backend using 'fastify-multer' lib

- [x] config multer from 'fastify-multer'
- [x] create plugin where add property file to request

savinaDaria:
- Added uploading plugin
- Tried to add preHandler, but it seems that it doesn't work
- Added 1 route 'POST' /api/v1/users/ for testing , then will delete it


nikita-remeslov:
@savinaDaria Provide an example of usage for upload!

savinaDaria:
@nikita-remeslov  This is the POST request I tried to use with sending file
`curl --location 'http://localhost:3000/api/v1/users/' \
--form 'file=@"/C:/Даша/1.txt"'`
Endpoint is defined in [user.controller.ts](https://github.com/BinaryStudioAcademy/bsa-2023-bsa-talents/pull/105/files#diff-dc598f09e8ea3c1c59e9dce410957a7c18ca4e29928e35a25dd493f91196b29d)

nikita-remeslov:
@savinaDaria still blocked?

savinaDaria:
@nikita-remeslov yeah, I also tried to change preHandler to another custom function instead of uploadFile.single('file') but it doesn't go there so there is problem with defining preHandler, I believe
I'll try to do something with it again in 2 hours

savinaDaria:
@nikita-remeslov 
I found a layer where it is saved. The main problem is that in the controller route we have a handler which accepts **options** parameter with the structure not **(request, reply)**, but this one **(body,query,params)** that is made  in controller.base.package.ts in mapRequest() method 
```
type ApiHandlerOptions<
    T extends DefaultApiHandlerOptions = DefaultApiHandlerOptions,
> = {
    body: T['body'];
    query: T['query'];
    params: T['params'];
};
```
And preHandler hook **uploadFile.single('file')** saves parsed file to field request.file in documentation
So I tested different parts and found out that for example in controller.base.package.ts in mapRequest() method this uploading could be done. 
image

So then I have question where should I save it, in body.file ?

Simpsonkk:
I think we also can use field limits in const uploadFile to limit file size and filefilter field to limit file extension (pdf, doc, docx)

savinaDaria:
USAGE Example: 

1.  In user.controller.ts add:

import { uploadFile } from '~/common/plugins/plugins.js';
...
this.addRoute({
            path: UsersApiPath.ROOT,
            method: 'POST',
            **preHandler: uploadFile.single('file'),**
            handler: (options) =>
                this.upload(options as ApiHandlerOptions<{
                            body: any;
                        }>,)
        });
 ...
        
private upload(
        options: ApiHandlerOptions<{
            body: any;
        }>,
    ):ApiHandlerResponse {
        **const uploadedFile = options.body.file;**
        console.log(uploadedFile)
        return {
            status: 200,
            payload: {},
        };
    }

image

backend/src/common/plugins/file-upload.plugin.ts
```diff
@@ -0,0 +1,56 @@
...
+        'fieldname' in file &&
+        'originalname' in file &&
+        'encoding' in file &&
+        'mimetype' in file
+    );
+}
+
+const fileFilter = (
+    request: FastifyRequest,
+    file: unknown | uploadedFile,
```
JosuerBague: consider using File type here?
```
import { type File as MulterFile, type FileFilterCallback} from 'fastify-multer/lib/interfaces.js'

const fileFilter = (request: FastifyRequest, file: MulterFile, cb: FileFilterCallback ): void => {} 

Removes the need to call isUploadedFile
```

backend/src/common/plugins/file-upload.plugin.ts
```diff
@@ -0,0 +1,56 @@
...
+            }
+        } else {
+            callback(new Error('Invalid file mimetype'), false);
+        }
+    }
+};
+
+const uploadFile = multer({
+    storage,
+    fileFilter: fileFilter,
```
JosuerBague:
```suggestion
    fileFilter,
```

backend/src/common/plugins/file-upload.plugin.ts
JosuerBague, line 1: I'm sorry I didn't notice this earlier. Maybe we should put this file inside a 'file-upload' dir:
```
backend/src/common/plugins/file-upload/{enums, file-upload.plugin.ts}
```
Since bt-21 will introduce an auth plugin.

JosuerBague (changes requested):

backend/src/common/plugins/file-upload/file-upload.plugin.ts
```diff
@@ -0,0 +1,50 @@
...
+
+const fileFilter = (
+    request: FastifyRequest,
+    file: MulterFile,
+    callback: FileFilterCallback,
+): void => {
+    if (
+        file.mimetype === AllowedMimeTypes.PDF ||
+        file.mimetype === AllowedMimeTypes.DOC ||
+        file.mimetype === AllowedMimeTypes.DOCX
```
MaksymGavatuk: Expand file types because we also need to upload image for avatar

backend/src/common/controller/controller-base.package.ts
```diff
@@ -49,8 +51,17 @@ class ControllerBase implements Controller {
     private mapRequest(
         request: Parameters<ServerAppRouteParameters['handler']>[0],
     ): ApiHandlerOptions {
-        const { body, query, params } = request;
-
+        const requestWithFile = request as FastifyRequest & {
+            file: MulterFile | undefined;
```
MaksymGavatuk:
```suggestion
            file?: MulterFile;
```

MaksymGavatuk (changes requested):

backend/src/common/plugins/file-upload/file-upload.plugin.ts
savinaDaria: @MaksymGavatuk  Ok, but I have a question, should I always save the file in 'file' field or separate fieldnames for images and for resume docs for example
So that regarding to the fieldname I will check mimetypes and extensions.
```
Example:
1. Endpoint has preHandler: uploadFile.single('document'). Then check if fieldname 'document' - check for pdf,doc,docx
2. Endpoint has preHandler: uploadFile.single('image'). Then check if fieldname 'image' - check for jpg,png

```
Or should I just add additional checks to if() and that's all?

savinaDaria: @MaksymGavatuk I guess I'll do separation for fieldnames but the result always will be in the body.file

savinaDaria:
Updated **Usage Example:**

In user.controller.ts add:
import { uploadFile } from '~/common/plugins/plugins.js';
...
this.addRoute({
path: UsersApiPath.ROOT,
method: 'POST',
**preHandler: uploadFile.single('document'), // preHandler: uploadFile.single('image'),**
handler: (options) =>
this.upload(options as ApiHandlerOptions<{
body: any;
}>,)
});
...

private upload(
options: ApiHandlerOptions<{
body: any;
}>,
):ApiHandlerResponse {
const uploadedFile = options.body.file;
console.log(uploadedFile)
return {
status: 200,
payload: {},
};
}

**CURL example:**

curl --location 'http://localhost:3000/api/v1/users/' \
--form 'document=@"/C:/Users/daria/OneDrive/Docs/resume.docx"'

curl --location 'http://localhost:3000/api/v1/users/' \
--form 'image=@"/C:/Users/daria/OneDrive/Docs/img.png"'

JosuerBague (approved):

backend/src/common/plugins/file-upload/file-upload.plugin.ts
```diff
@@ -0,0 +1,75 @@
...
+            ].includes(file.mimetype);
+            isValidExtension = [
+                AllowedExtensions.JPEG,
+                AllowedExtensions.JPG,
+                AllowedExtensions.PNG,
+            ].includes(fileExtension);
+            break;
+        }
+        default: {
+            callback(new Error('Invalid file group'), false);
```
nikita-remeslov: add this error to shared

backend/src/common/plugins/file-upload/file-upload.plugin.ts
```diff
@@ -0,0 +1,75 @@
...
+            return;
+        }
+    }
+
+    if (isValidMimeType && isValidExtension) {
+        callback(null, true);
+    } else {
+        callback(new Error(`Invalid ${file.fieldname} file`), false);
```
nikita-remeslov: also, we should add to shared.

nikita-remeslov (changes requested):

backend/src/common/plugins/file-upload/file-upload.plugin.ts
MaksymGavatuk: You don't need to separate types

MaksymGavatuk (approved):

anton-otroshchenko (approved):

nikita-remeslov (approved):

PR #106, opened 2023-08-25
bt-91: create slider component

### LINKED ISSUE #91: Create Slider component
Research existing libraries (and discuss them with team) for the Slider component. It can be a bit different from the UI but the main idea should be the same. It also should be a shared component with custom data provided to it 

Image

mskmee:
**Dont forget to rebuild shared**

Used string color variable in slider, because we got the thinnest line and color from Figma didn't work quite well. 
Did'n add it to colors, because it's the only thing where we will use it.
slider2
slider


anogire:
Will it look like this? 
image


Can we make it wider, like in Figma, or is it not possible?

mskmee:
> Will it look like this? image
> 
> Can we make it wider, like in Figma, or is it not possible?

Library didn't let us do it with default. But we can make it more similar to figma, component will grow up and make it more complicated. Katerina said leave it like this for now.

mobile/src/bundles/common/components/slider/slider.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+import { styles } from './styles';
+
+type SliderProperties = {
+    style?: StyleProp<ViewStyle>;
+};
+
+const minSliderValue = 0;
+const maxSliderValue = 120;
+const valueWidth = 70;
```
katerynakostikova: These values should be passed via props. Let's add default component props to SliderProperties

mobile/src/bundles/common/components/slider/slider.tsx
```diff
@@ -0,0 +1,64 @@
...
+import { Pressable, Text, View } from '~/bundles/common/components/components';
+import { Color, TextCategory } from '~/bundles/common/enums/enums';
+import { convertInputToDate } from '~/bundles/common/helpers/helpers';
+import { useCallback, useWindowDimensions } from '~/bundles/common/hooks/hooks';
+import { globalStyles } from '~/bundles/common/styles/styles';
+
+import { styles } from './styles';
+
+type SliderProperties = {
+    style?: StyleProp<ViewStyle>;
```
katerynakostikova:
```suggestion
    containerStyle?: StyleProp<ViewStyle>;
```

mobile/src/bundles/common/components/slider/slider.tsx
```diff
@@ -0,0 +1,64 @@
...
+    return (
+        <View style={[globalStyles.p5, viewStyle]}>
+            <View style={[globalStyles.flexDirectionRow]}>
+                <Text category={TextCategory.LABEL} style={styles.label}>
+                    Experience
+                </Text>
+                <Text category={TextCategory.LABEL} style={styles.label_icon}>
+                    *
+                </Text>
+            </View>
```
katerynakostikova: Firstly, we have a FormField component already with this logic. Secondly, this component should only represent data passed via props, so that we can use it in multiple places. So for now you don't need any date functions and so on, just representing provided data

katerynakostikova: So there should be basic slider props, and some values under the slider, like"no", which also should be passed via props

shared/src/helpers/convert-input-to-date/convert-input-to-date.ts
```diff
@@ -0,0 +1,15 @@
...
+    const monthInYear = 12;
+    const yearRounder = 10;
+    const dateTitle = value < monthInYear ? 'month' : 'year';
+    const dateValue =
+        value < monthInYear
+            ? value
+            : Math.round((value / monthInYear) * yearRounder) / yearRounder;
+    const refactoredDateTitle = getDateEnding(dateTitle, dateValue);
+    return `${dateValue} ${refactoredDateTitle}`;
+};
```
katerynakostikova: For now, we don't know if there will be a possibility to add a month, you can keep this function somewhere if you don't want to lose it, but for now, we don't need it and it's better to remove it. We'll also use day js lib to work with dates so it may be easier in future

illia-dobro (commented):
align
Perhaps adjust alignment here.
And it would be better if, at the end of the line (on the right), we see "10+ years". If it is possible

mskmee:
It's possible. But like Katerina write this component should only representative data, so that's value should come from props. As we will know what values shall we have, we will implement it)

mobile/src/bundles/common/components/slider/slider.tsx
mskmee: fix

mobile/src/bundles/common/components/slider/slider.tsx
mskmee: fix

mobile/src/bundles/common/components/slider/slider.tsx
mskmee: fix

shared/src/helpers/convert-input-to-date/convert-input-to-date.ts
mskmee: fix

mobile/src/bundles/common/components/slider/slider.tsx
```diff
@@ -0,0 +1,68 @@
+import CommunitySlider from '@react-native-community/slider';
```
katerynakostikova:
```suggestion
import CommunitySlider, { type SliderProps } from '@react-native-community/slider';
```

mobile/src/bundles/common/components/slider/slider.tsx
```diff
@@ -0,0 +1,68 @@
...
+type SliderProperties = {
+    thumbTitleValue: string;
+    sliderValue: number;
+    onSliderValueChange: (value: number) => void;
+    onResetClickHandler: () => void;
+    containerStyle?: StyleProp<ViewStyle>;
+    minSliderValue?: number;
+    maxSliderValue?: number;
+    thumbTitleValueWidth?: number;
+    resetTitle?: string;
```
katerynakostikova: This should be just initial value (min value), for example 0 which also means no

mobile/src/bundles/common/components/slider/slider.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+type SliderProperties = {
+    thumbTitleValue: string;
+    sliderValue: number;
+    onSliderValueChange: (value: number) => void;
+    onResetClickHandler: () => void;
+    containerStyle?: StyleProp<ViewStyle>;
+    minSliderValue?: number;
+    maxSliderValue?: number;
+    thumbTitleValueWidth?: number;
```
katerynakostikova:
```suggestion
type SliderProperties = SliderProps & {
    thumbTitleValue: string;
    containerStyle?: StyleProp<ViewStyle>;
```
Let's keep only custom props here

mskmee: Good morning, but it shall destroy abstact layer for the library or it's okey for us?

mobile/src/bundles/common/components/slider/slider.tsx
mskmee: Didn't understand it, we need min value for slider width, active value, and set value we got from parent component and return it to actual for form controller.
Use thumbTitleValueWidth for calculation offset and set it to text component. Maybe we should use a better way for it?)
Noted that I should remove reset title and so on, because this component didn't need to know about them)

katerynakostikova: My comment was about resetTitle, GitHub displayed it wrong

mobile/src/bundles/common/components/slider/slider.tsx
katerynakostikova: What do you mean

mskmee: fix

mobile/src/bundles/common/components/slider/slider.tsx
mskmee: fix

mobile/src/bundles/common/components/slider/slider.tsx
mskmee: fix

mobile/src/bundles/common/components/slider/slider.tsx
```diff
@@ -0,0 +1,60 @@
...
+import { useWindowDimensions } from '~/bundles/common/hooks/hooks';
+import { globalStyles } from '~/bundles/common/styles/styles';
+
+import { styles } from './styles';
+
+type SliderProperties = {
+    thumbTitleValue: string;
+    containerStyle?: StyleProp<ViewStyle>;
+    value: number;
+    thumbTitleValueWidth?: number;
```
katerynakostikova: Please put optional always as last 
```suggestion
    thumbTitleValue: string;
     value: number;
     thumbTitleValueWidth?: number;
    containerStyle?: StyleProp<ViewStyle>;
```

mobile/src/bundles/common/components/slider/slider.tsx
```diff
@@ -0,0 +1,60 @@
...
+} & Omit<SliderProps, 'style' | 'value'>;
+
+const defaultMinSliderValue = 0;
+const defaultMaxSliderValue = 120;
+const defaultValueWidth = 70;
```
katerynakostikova: What is default value?

mskmee: Value is required porps or it bad practice?

katerynakostikova: Ok, let's make them more usual, for example from 0 to 10 and 0 as default

mskmee: fix

mobile/src/bundles/common/components/slider/slider.tsx
mskmee: fix

katerynakostikova (approved):

anogire (approved):

illia-dobro (approved):

nikita-remeslov (approved):

PR #114, opened 2023-08-25
bt-53: Create sidebar

### LINKED ISSUE #53: Create Sidebar
Image

Design: https://www.figma.com/file/Yd7USgP7PLvA0W794jHXWu/BSA_recruting?type=design&node-id=0-1&mode=design&t=lwbtBxp8JzQcOuO5-0

vladjsp:
sidebar2


JemiJe:
I saw prev commit, but maybe we should leave ability to pass `menuItems` via props, but leave default items and add passed using, for example `menuItems = menuItems.concat(extraItems)`. As I see it, better to make component more extensible.

savinaDaria (approved):

frontend/src/bundles/common/components/sidebar/sidebar.tsx
```diff
@@ -0,0 +1,43 @@
...
+        link: AppRoute.SIGN_UP,
+        name: 'Chats',
+        icon: <EmailRounded />,
+    },
+];
+
+const Sidebar: React.FC = () => {
+    return (
+        <div className={styles.wrapper}>
+            <div className={styles.logo}></div>
```
nikita-remeslov: Leave some comment, that we should update it after we will have our logo

frontend/src/bundles/common/components/sidebar/sidebar.tsx
nikita-remeslov, line 1: you should import it to the root components.ts file

nikita-remeslov (changes requested):
In general looks good

Simpsonkk:
Probably we can also use z-index: 2 in .wrapper and position: fixed because the sidebar is always located above the header component.

nikita-remeslov (approved):

PR #119, opened 2023-08-26
bt-104: Create complete profile screen UI

### LINKED ISSUE #104: Create Complete Profile screen UI
Create Profile screen UI. 

Image

anogire:
**isProfileComplete** must be set to **true**

```
const Root: React.FC<Properties> = ({
    isSignedIn = false,
    isProfileComplete = true,
}) => {
```

At the moment, it looks like this.
Screenshot 2023-08-26 at 19 52 52


mobile/src/bundles/common/components/currency-input/currency-input.tsx
likeri29, line 1: I think that I have to remake Input component and conditionally add this dollar sign or smth else. Because we also have this input: 
image

mobile/src/bundles/common/enums/navigation/onboarding-screen-name.enum.ts
```diff
@@ -1,6 +1,9 @@
 const OnboardingScreenName = {
     PROFILE: 'Profile',
     BSA_BADGES: 'BSA Badges',
+    SKILLS_PROJECTS: 'Skills and projects',
+    CV_CONTACTS: 'CV and contacts',
+    PREVIEW: 'Preview',
```
likeri29, line 6: We will have some merging conflicts because I've also created these screens.

mobile/src/bundles/common/components/currency-input/currency-input.tsx
anogire: Maybe yes. Make a separate DoubleInput and pass 'label' as props: '$', and so on.

mobile/src/bundles/common/enums/navigation/onboarding-screen-name.enum.ts
anogire: I only added the names for the automatic calculation of the number of steps in STEP 01/05. I think this conflict will be easy to resolve later.

likeri29: Yeah

anogire:
1. I think we need to make a separate CheckboxGroup component (react-native-paper library).

2. Currently, the Selector does not show "*" as a required field because it has its own 'Label', but we need the 'Label' from the FormField to be displayed.
Also, the values overlap with other elements. 
It seems to me that it is better to fix the styles of the Selector itself.

Screenshot 2023-08-27 at 01 07 15
Screenshot 2023-08-27 at 01 07 51

Screenshot 2023-08-27 at 01 08 28
Screenshot 2023-08-27 at 01 08 58

katerynakostikova:
Let's use https://github.com/react-native-checkbox/react-native-checkbox for checkbox

katerynakostikova:
For selector there should be background for list of items, let's fix styles and remove label

mobile/src/bundles/common/components/input/input.tsx
```diff
@@ -10,39 +10,69 @@ import { TextInput } from 'react-native';
...
+import { Text, View } from '../components';
 import { styles } from './styles';
 
 type Properties<T extends FieldValues> = TextInputProps & {
     control: Control<T, null>;
     name: FieldPath<T>;
     hasError?: boolean;
+    marker?: string;
+    inputType?: 'text' | 'number';
+    multilines?: boolean;
```
katerynakostikova: Some of this should already be in TextInputProps

mobile/src/bundles/common/components/input/styles.ts
```diff
@@ -3,9 +3,19 @@ import { StyleSheet } from 'react-native';
...
+        borderTopLeftRadius: 5,
+        borderBottomLeftRadius: 5,
+    },
     input: {
         borderWidth: 1,
-        borderColor: '#7e7e7e',
+        borderColor: Color.INPUT,
+        borderTopRightRadius: 5,
+        borderBottomRightRadius: 5,
+    },
```
katerynakostikova, line 16: Does this style work correctly if there is no marker?

mobile/src/bundles/users/components/profile-form/profile-form.tsx
```diff
@@ -0,0 +1,129 @@
...
+        { label: 'Location 2', value: 'location2' },
+        { label: 'Location 3', value: 'location3' },
+    ];
+
+    const handleFormSubmit = useCallback((): void => {
+        void handleSubmit(onSubmit)();
+    }, [handleSubmit, onSubmit]);
+
+    return (
+        <FlatList
```
katerynakostikova: Let's use ScrollView since we know that it's not an endless or long list

anogire: Selector component uses FlatList and there is a problem with nested scrolling in ScrollView.
Because of this, I switched to FlatList too.

mobile/src/bundles/common/components/input/styles.ts
anogire: Yes, because in <TextInput> I use 
`!marker && globalStyles.borderRadius5,`
for this

mobile/src/bundles/common/components/input/input.tsx
anogire: Oh, yes, we have 'multiline' already.

anogire:
This is for     
```
dropdown: {
        position: 'absolute',
        top: '100%',
         ...
    },
```
Screenshot 2023-08-27 at 10 06 50


And without "top: '100%'"
```
dropdown: {
        position: 'absolute',
         ...
    },
```
Screenshot 2023-08-27 at 10 07 08

mobile/src/bundles/users/components/profile-form/profile-form.tsx
```diff
@@ -0,0 +1,128 @@
...
+                            multiline={true}
+                        />
+                    </FormField>
+                </>
+            }
+            ListFooterComponent={
+                <View style={globalStyles.flexDirectionRow}>
+                    <Button label="Back" disabled style={globalStyles.mr10} />
+                    <Button label="Next" onPress={handleFormSubmit} />
+                </View>
```
illia-dobro, line 143: These buttons appear on other screens (I have them too in my task). Perhaps we could create a separate component for them.

anogire: I'm not sure we need a separate component. We can pass the necessary styles to the buttons

anogire:
I like this look of Selector better, what do you think? 
The "arrow" overlaps, so you can't close it without selecting something. 
But maybe that's the way it should be?
Screenshot 2023-08-27 at 17 24 33

The view at the moment
Screenshot 2023-08-27 at 17 25 35

mobile/src/bundles/common/components/checkbox/checkbox.tsx
```diff
@@ -0,0 +1,54 @@
...
+            style={[
+                globalStyles.flexDirectionRow,
+                globalStyles.alignItemsFlexStart,
+                styles.container,
+            ]}
+        >
+            <CheckBox
+                value={value}
+                onValueChange={onChange}
+                tintColors={{ true: Color.PRIMARY, false: Color.INPUT }}
```
illia-dobro, line 43: I used your checkbox component. It works, but the "disable"  does not work. Maybe it should be written like this: const colorCheckbox = props.disabled ? Color.TEXT2 : Color.PRIMARY;

tintColors={{ true: colorCheckbox, false: Color.INPUT }}

anogire: I've been testing "disabled."
Just when you call a component you can write 'disabled' and this property will be passed.
It is his own prop.
Like:
<Checkbox ... disabled> and it will be disabled.

illia-dobro:
> I like this look of Selector better, what do you think? The "arrow" overlaps, so you can't close it without selecting something. But maybe that's the way it should be? Screenshot 2023-08-27 at 17 24 33

I've tried some changes:
selector

mobile/src/bundles/common/components/checkbox/checkbox.tsx
illia-dobro: But color does not change

illia-dobro: check

anogire: Well, as far as I understand.
If the checkbox is disabled, it means that it cannot be "checked". 
So it will look like this and will not respond to clicks.
Screenshot 2023-08-27 at 20 51 33

likeri29:
> I've tried some changes: selector

I think that this variant is better, but is it possible to add smth like line (add border bottom) to divide possible options? It will be  more visually noticeable

anogire:
Maybe so.
But borderBottom after each item seems to me to be too much.
Screenshot 2023-08-27 at 21 59 22

likeri29:
> Maybe so.
> 
> But borderBottom after each item seems to me to be too much.
> 
> Screenshot 2023-08-27 at 21 59 22
> 
> 

I meant like this:image

Because if we have more than 3 options it will be like blank space with a bunch of options which won't be separated.

anogire:
Yes, I understand. 
But in my opinion, there will be too many "lines". In our small screen on the phone....

mobile/src/bundles/users/components/new-account-header/new-account-header.tsx
```diff
@@ -0,0 +1,44 @@
...
+    OnboardingScreenName,
+    TextCategory,
+} from '~/bundles/common/enums/enums';
+import { globalStyles } from '~/bundles/common/styles/styles';
+
+import { StepProgressBar } from './step-progress-bar';
+import { StepTrack } from './step-track';
+
+type Properties = {
+    title: (typeof OnboardingScreenName)[keyof typeof OnboardingScreenName];
```
MaksymGavatuk: Can we use ValueOf in this case ?

MaksymGavatuk (changes requested):

mobile/src/bundles/users/components/new-account-header/new-account-header.tsx
anogire: Yes ... I didn't notice that we have already such type ... 
I will fix it, thank you

mobile/src/bundles/users/components/profile-form/profile-form.tsx
katerynakostikova: Let's try to fix the selector somehow

anogire: Then I suggest the following implementation of the Selector component: [commit](https://github.com/BinaryStudioAcademy/bsa-2023-bsa-talents/pull/119/commits/744f7a3593621c6c249d2640f3d688b2f7afde1f)

mobile/src/bundles/common/components/checkbox/checkbox.tsx
katerynakostikova: Looks ok for disabled

mobile/src/bundles/common/components/input/input.tsx
```diff
@@ -7,42 +7,72 @@ import {
...
+    inputType?: 'text' | 'number';
 };
 
 const Input = <T extends FieldValues>({
     editable = true,
     hasError,
     control,
     name,
+    marker,
+    inputType = 'text',
```
katerynakostikova: we can pass just keyboard type

mobile/src/bundles/common/components/selector/selector.tsx
```diff
@@ -1,16 +1,15 @@
-import { type ReactElement, useState } from 'react';
+import { useState } from 'react';
```
katerynakostikova:
```suggestion
import  React, { useState } from 'react';
```

mobile/src/bundles/common/components/selector/selector.tsx
```diff
@@ -57,46 +54,40 @@ const Selector: React.FC<Properties> = ({ label, options }) => {
...
+                        >
+                            <Text
+                                category={TextCategory.LABEL}
+                                style={globalStyles.pv5}
+                            >
+                                {item.label}
+                            </Text>
+                        </TouchableOpacity>
+                    ))}
+                </View>
```
katerynakostikova: What if this selector has a lot of items, 20-30 for example, I guess we still need flatList, but maxHeight of the list should be defined

mobile/src/bundles/common/components/selector/styles.ts
```diff
@@ -1,16 +1,17 @@
...
         borderColor: Color.INPUT,
     },
     dropdown: {
-        position: 'absolute',
+        marginTop: -5,
+        borderBottomRightRadius: 5,
+        borderBottomLeftRadius: 5,
         backgroundColor: '#FFF',
-        zIndex: 1,
     },
```
katerynakostikova: What if under the toggled list there is an input for example? This list should appear upon it and it shouldn't move layout

mobile/src/bundles/users/components/new-account-header/new-account-header.tsx
```diff
@@ -0,0 +1,45 @@
...
+    return (
+        <>
+            <View
+                style={[
+                    globalStyles.flexDirectionRow,
+                    globalStyles.justifyContentSpaceBetween,
+                    globalStyles.alignItemsCenter,
+                    globalStyles.borderRadiusTop8,
+                    globalStyles.p20,
+                    { backgroundColor: '#FFFFFF' },
```
katerynakostikova:
```suggestion
                    { backgroundColor: '#FFFFFF' },
```
Let's move to styles

mobile/src/bundles/users/components/new-account-header/step-progress-bar.tsx
```diff
@@ -0,0 +1,30 @@
...
+type Properties = {
+    currentStep: number;
+    totalSteps: number;
+};
+
+const StepProgressBar: React.FC<Properties> = ({ currentStep, totalSteps }) => {
+    const max = 100;
+
+    return (
+        <View style={[globalStyles.flexDirectionRow, { height: 3 }]}>
```
katerynakostikova: { height: 3 } also to styles

mobile/src/bundles/users/components/new-account-header/step-progress-bar.tsx
```diff
@@ -0,0 +1,30 @@
...
+            <View
+                style={{
+                    width: `${(max / totalSteps) * currentStep}%`,
+                    backgroundColor: Color.PRIMARY,
+                }}
+            />
+            <View
+                style={[globalStyles.width100, { backgroundColor: Color.TAG }]}
```
katerynakostikova: same here, inline styles only acceptable if they depend on some props

mobile/src/bundles/users/components/new-account-header/step-track.tsx
```diff
@@ -0,0 +1,34 @@
...
+import { globalStyles } from '~/bundles/common/styles/styles';
+
+type Properties = {
+    currentStep: number;
+    totalSteps: number;
+};
+
+const formatStep = (step: number, width: number): string => {
+    return String(step).padStart(width, '0');
+};
```
katerynakostikova: let's move this to helpers

mobile/src/bundles/users/components/new-account-header/step-track.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+const StepTrack: React.FC<Properties> = ({ currentStep, totalSteps }) => {
+    const stepWidth = 2;
```
katerynakostikova: can be moved out of component, so it doesn't recreates on re-render 
```suggestion
    const stepWidth = 2;
```

mobile/src/bundles/users/components/new-account-header/step-track.tsx
```diff
@@ -0,0 +1,34 @@
...
+    const paddedStep = formatStep(currentStep, stepWidth);
+    const paddedTotalSteps = formatStep(totalSteps, stepWidth);
+
+    return (
+        <View style={globalStyles.flexDirectionRow}>
+            <Text category={TextCategory.STEP} style={{ letterSpacing: 1.5 }}>
```
katerynakostikova: also styles

mobile/src/bundles/users/components/profile-form/constants/constants.ts
```diff
@@ -0,0 +1,54 @@
...
+        fullTime: true,
+        partTime: false,
+        freelance: false,
+        partTime2: false,
+        remotely: false,
+        relocation: false,
+    },
+};
+
+const JOB_TITLE_OPTIONS = [
```
katerynakostikova, line 19:
```suggestion
// TODO: remove 
const JOB_TITLE_OPTIONS = [
```

mobile/src/bundles/users/components/profile-form/constants/constants.ts
```diff
@@ -0,0 +1,54 @@
...
+    { label: 'Job Title 18', value: 'job18' },
+    { label: 'Job Title 19', value: 'job19' },
+    { label: 'Job Title 20', value: 'job20' },
+    { label: 'Job Title 21', value: 'job21' },
+    { label: 'Job Title 22', value: 'job22' },
+    { label: 'Job Title 23', value: 'job23' },
+    { label: 'Job Title 24', value: 'job24' },
+];
+
+const CURRENT_LOCATION_OPTIONS = [
```
katerynakostikova, line 47: same here

mobile/src/bundles/users/screens/profile.tsx
```diff
@@ -1,9 +1,27 @@
...
+    const { name } = useAppRoute();
+    const stepTitle = name as ValueOf<typeof OnboardingScreenName>;
+
+    const handleProfileSubmit = useCallback(() => {
+        return null;
+    }, []);
+
+    return (
+        <View style={[globalStyles.flex1, { backgroundColor: Color.TEXT }]}>
+            <NewAccountHeader title={stepTitle} currentStep={1} />
```
katerynakostikova:
```suggestion
            <NewAccountHeader title={stepTitle} currentStep={1} />
```
Let's add map for each screen step number

mobile/src/bundles/users/screens/profile.tsx
```diff
@@ -1,9 +1,27 @@
...
+
+import { NewAccountHeader, ProfileForm } from '../components/components';
 
 const Profile: React.FC = () => {
-    return <Text>Step one PROFILE</Text>;
+    const { name } = useAppRoute();
+    const stepTitle = name as ValueOf<typeof OnboardingScreenName>;
+
+    const handleProfileSubmit = useCallback(() => {
+        return null;
```
katerynakostikova, line 22:
```suggestion
        // TODO: add submit logic 
        return null;
```

anogire:
Screenshot 2023-08-28 at 19 00 21

anogire:
1. I made some small changes to the styles for the Slider component because it should fill the whole width by design.

2. And as far as I understand, later we will need to add **'name'** and **'control'** to the Slider and Selector components. Am I right?

Screenshot 2023-08-29 at 02 02 53

katerynakostikova:
> 2\. And as far as I understand, later we will need to add **'name'** and **'control'** to the Slider and Selector components. Am I right?

yes

mobile/src/bundles/common/components/checkbox/styles.tsx
```diff
@@ -0,0 +1,9 @@
+import { StyleSheet } from 'react-native';
+
+const styles = StyleSheet.create({
+    container: {
+        marginLeft: -7,
```
katerynakostikova: Why -7?

katerynakostikova: Let's use containerStyle prop for this and pass this if needed,

mobile/src/bundles/common/components/input/input.tsx
```diff
@@ -7,42 +7,68 @@ import {
...
+                        globalStyles.ph20,
+                        styles.marker,
+                    ]}
+                >
+                    <Text category={TextCategory.INPUT}>{marker}</Text>
+                </View>
+            )}
+            <TextInput
+                onChangeText={onChange}
+                value={value.toString()}
```
katerynakostikova: it's not a string by default?

mobile/src/bundles/common/components/slider/styles.ts
```diff
@@ -3,6 +3,7 @@ import { StyleSheet } from 'react-native';
 const styles = StyleSheet.create({
     slider: {
         borderRadius: 6,
+        marginHorizontal: -10,
```
katerynakostikova: again why -? And can we pass it via containerStyle?

mobile/src/bundles/users/components/profile-form/constants/constants.ts
```diff
@@ -0,0 +1,58 @@
+// TODO: remove
```
katerynakostikova:
```suggestion

```
I guess default values will be needed anyway

mobile/src/bundles/common/components/checkbox/styles.tsx
anogire: Because by default, the component has a default margin that I can't remove, and by design, it should be below the text on top.

mobile/src/bundles/common/components/slider/styles.ts
anogire: Here, it should be a style not for the View, but for the CommunitySlider inside, so it cannot be applied throw  containerStyle. Unless I apply containerStyle directly to the CommunitySlider.
Or we will change the design a bit.
Then it will be like this:
Screenshot 2023-08-29 at 11 44 25

mobile/src/bundles/common/components/input/input.tsx
anogire: Yes, it's a legacy from when I didn't pass my own keyboardType. I don't need it now.

mobile/src/bundles/common/components/checkbox/styles.tsx
anogire: I did it through containerStyle

mobile/src/bundles/common/components/slider/styles.ts
katerynakostikova: Let's keep everything without margin: -value for now, we'll fix it later if needed

mobile/src/bundles/common/components/selector/selector.tsx
```diff
@@ -44,8 +43,7 @@ const Selector: React.FC<Properties> = ({ label, options }) => {
         : IconName.ARROW_DROP_DOWN;
 
     return (
-        <View>
-            <Text category={TextCategory.LABEL}>{label}</Text>
+        <View style={{ position: 'relative' }}>
```
katerynakostikova: remove inline styles (move to styles.ts)

katerynakostikova (approved):

mobile/src/bundles/common/components/selector/selector.tsx
anogire: done

mobile/src/bundles/common/components/slider/styles.ts
anogire: done

anogire: Screenshot 2023-08-29 at 12 53 18

mobile/src/bundles/users/components/profile-form/constants/constants.ts
nikita-remeslov: It can be shared enum

mobile/src/bundles/users/components/profile-form/constants/constants.ts
nikita-remeslov: it can be shared enum

nikita-remeslov (approved):

mobile/src/bundles/users/components/new-account-header/new-account-header.tsx
anogire: done

illia-dobro (approved):

MaksymGavatuk (approved):

PR #122, opened 2023-08-27
bt-72: create user profile(files and badges)

### LINKED ISSUE #72: Create user profile
Look at screens (6-9) in Figma and create user profile schema and relations for profile creation

**Please follow to the next steps:**

- [x] create relations diagram (ER)
- [x] approve your diagram with coaches
- [x] create migrations for described tables

Image
Image
Image
Image

nika-sirak:
Added user_email to talent_badges table and removed user_id, because when the admin creates talent_badges, the badge should have a field that is linked to a specific user, id doesn't exist yet, but the email is already known. 

If a user decides to leave our application, their badges should not be deleted from the database, because this information is needed for the academy. 
Do we need to link bsa-badges with the users table in this case?
Could you, please, clarify about types of FK?  They all should be string or integer?

Screenshot 2023-08-27 at 12 32 12


backend/src/migrations/20230827061743_add_talent_badges_table.ts
```diff
@@ -0,0 +1,52 @@
...
+    UPDATED_AT: 'updated_at',
+};
+
+const RelationRule = {
+    CASCADE: 'CASCADE',
+} as const;
+
+async function up(knex: Knex): Promise<void> {
+    return knex.schema.createTable(TableName.TALENT_BADGES, (table) => {
+        table.increments(ColumnName.ID).primary();
```
MaksymGavatuk: const uuid = 'uuid_generate_v4()';
...
table
  .uuid('id')
  .unique()
  .notNullable()
  .defaultTo(knex.raw(uuid))
  .primary({ constraintName: 'hard_skills_pkey' });

MaksymGavatuk (changes requested):

backend/src/migrations/20230826220746_add_bsa_badges_table.ts
```diff
@@ -0,0 +1,43 @@
...
+    TYPE: 'type',
+    NAME: 'name',
+    MAX_SCORE: 'max_score',
+    CREATED_AT: 'created_at',
+    UPDATED_AT: 'updated_at',
+};
+
+async function up(knex: Knex): Promise<void> {
+    return knex.schema.createTable(TABLE_NAME, (table) => {
+        table
```
MaksymGavatuk: When you use uuid function first of all you need to create extension if not exists, because in other db this migration can fall down without the necessary extension. Every migration need to has this line when you use uuid

```suggestion
async function up(knex: Knex): Promise<void> {
    await knex.raw('CREATE EXTENSION IF NOT EXISTS "uuid-ossp";');

    return knex.schema.createTable(TABLE_NAME, (table) => {
        table
```

backend/src/migrations/20230826223636_add_role_to_users_table.ts
```diff
@@ -0,0 +1,23 @@
+import { type Knex } from 'knex';
```
MaksymGavatuk: Name migrations like:

`add_${tableName}_table.ts` - to create table
`update_${tableName}_table.ts` - to update table

backend/src/migrations/20230827061743_add_talent_badges_table.ts
```diff
@@ -0,0 +1,60 @@
...
+    UPDATED_AT: 'updated_at',
+};
+
+const RelationRule = {
+    CASCADE: 'CASCADE',
+} as const;
+
+async function up(knex: Knex): Promise<void> {
+    return knex.schema.createTable(TableName.TALENT_BADGES, (table) => {
+        table
```
MaksymGavatuk: The same
```suggestion
async function up(knex: Knex): Promise<void> {
    await knex.raw('CREATE EXTENSION IF NOT EXISTS "uuid-ossp";');

    return knex.schema.createTable(TableName.TALENT_BADGES, (table) => {
        table
```

backend/src/migrations/20230828083409_update_id_users_table.ts
```diff
@@ -0,0 +1,39 @@
...
+    ID: 'id',
+};
+
+async function up(knex: Knex): Promise<void> {
+    await knex.schema.alterTable(TABLE_NAME, (table) => {
+        table.dropPrimary();
+        table.dropColumn(ColumnName.ID);
+    });
+
+    return knex.schema.alterTable(TABLE_NAME, (table) => {
```
MaksymGavatuk: The same
```suggestion
    await knex.raw('CREATE EXTENSION IF NOT EXISTS "uuid-ossp";');

    return knex.schema.alterTable(TABLE_NAME, (table) => {
```

MaksymGavatuk (changes requested):

nikita-remeslov:
@nika-sirak you should merge development to this branch `task/bt-72-create-user-profile`

nikita-remeslov (approved):
Looks good to me

MaksymGavatuk (approved):

PR #135, opened 2023-08-28
bt-133: Disable some linter rules for d.ts files in frontend

### LINKED ISSUE #133: Disable some linter rules for d.ts files in frontend
frontend/src/bundles/common/themes/typography-theme/typography.d.ts
frontend/src/bundles/common/themes/color-theme/theme.d.ts

disable the linter rule for these files and remove comments in files.

nika-sirak:
Disabled the linter rule for `typography.d.ts` and `theme.d.ts` files and removed comments in these files.

### DIFF
--- a/frontend/.eslintrc.yml
+++ b/frontend/.eslintrc.yml
@@ -31,6 +31,13 @@ overrides:
               - off
     - files:
           - vite-env.d.ts
+          - typography.d.ts
       rules:
           unicorn/prevent-abbreviations:
+              - off
+    - files:
+          - typography.d.ts
+          - theme.d.ts
+      rules:
+          '@typescript-eslint/consistent-type-definitions':
               - off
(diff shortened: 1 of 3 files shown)


frontend/.eslintrc.yml
```diff
@@ -31,6 +31,13 @@ overrides:
               - off
     - files:
           - vite-env.d.ts
+          - typography.d.ts
       rules:
           unicorn/prevent-abbreviations:
               - off
+    - files:
+          - typography.d.ts
```
nikita-remeslov: Looks like it can be like this
```
- files: 
       - *.d.ts
- rules:
       // all rules

nikita-remeslov (changes requested):

frontend/.eslintrc.yml
nika-sirak: thx for this suggestion

nikita-remeslov (approved):

mskmee (approved):

PR #137, opened 2023-08-28
bt-127: Create shared Modal component

### LINKED ISSUE #127: Create shared Modal component
Create a shared Modal component. It should have a Material UI design style. It should be a window that can be closed on the backdrop and we can pass any options there.

likeri29:
[
Modal
](url)

Example of usage:
example


mobile/src/bundles/common/components/modal/modal.tsx
```diff
@@ -0,0 +1,82 @@
...
+    View,
+} from '~/bundles/common/components/components';
+import { ButtonType, TextCategory } from '~/bundles/common/enums/enums';
+import { globalStyles } from '~/bundles/common/styles/styles';
+
+import { styles } from './styles';
+
+type Properties = {
+    title: string;
+    visible: boolean;
```
mskmee: boolean variables supposte to start from is, are, have, had and so on

mobile/src/bundles/common/components/modal/modal.tsx
```diff
@@ -0,0 +1,82 @@
...
+                                label={acceptText}
+                                onPress={onAccept}
+                            />
+                        )}
+                    </View>
+                </View>
+            </TouchableOpacity>
+        </RNModal>
+    );
+};
```
mskmee: One of usages is choosing image picker source, think it's better just receive children inside with logic from parent. Need Katerina opinion.

mobile/src/bundles/common/components/modal/styles.ts
```diff
@@ -0,0 +1,19 @@
+import { StyleSheet } from 'react-native';
+
+// import { Color } from '~/bundles/common/enums/enums';
```
mskmee: commented code

mobile/src/bundles/common/components/modal/modal.tsx
```diff
@@ -0,0 +1,82 @@
+import React from 'react';
+import { Modal as RNModal } from 'react-native';
```
mskmee: guess we should import it from `~/bundles/common/components/components`

likeri29: When Im doing this we have like conflict, because of to Modals in the same file

mobile/src/bundles/common/components/modal/modal.tsx
likeri29: This is the Modal's property

mobile/src/bundles/common/components/modal/modal.tsx
katerynakostikova: Yes, let's create some basic modal with children and this option should be called ActionModal (which uses shared basic Modal) for example

likeri29: To be honest I do not fully understand what I should do. Should I create one more component which will just accept only children and no other props?

katerynakostikova: Yes, because we need something like this in future and that Modal with buttons is also needed for us 
image

likeri29: I've created simple modal which accepts 3 props (children, onClose (to close the modal by clicking outside), and visible)

mskmee (approved):

mobile/src/bundles/common/components/active-modal/active-modal.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+import { styles } from './styles';
+
+type Properties = {
+    visible: boolean;
+    children: React.ReactNode;
+    onClose: () => void;
+};
+
+const ActiveModal: React.FC<Properties> = ({ visible, children, onClose }) => {
```
katerynakostikova: This component should be called "Modal" (it's just wrapper)

mobile/src/bundles/common/components/modal/modal.tsx
```diff
@@ -0,0 +1,82 @@
...
+type Properties = {
+    title: string;
+    visible: boolean;
+    acceptText?: string;
+    children: React.ReactNode;
+    onClose: () => void;
+    onAccept?: () => void;
+};
+
+const Modal: React.FC<Properties> = ({
```
katerynakostikova: And this one should be called "ActionModal" (it requires some action from user, because it has buttons)

mobile/src/bundles/common/components/modal/styles.ts
```diff
@@ -0,0 +1,17 @@
...
+    },
+    container: {
+        width: '80%',
+        gap: 20,
+        backgroundColor: '#fff',
+    },
+    content: {
+        gap: 10,
+    },
+});
```
katerynakostikova: Let's create these 2 components in one folder cause they are similar and have same styles 
```
--common
    --components
       --modals
           --moldal.tsx
           --action-moldal.tsx
           --styles.ts
```

katerynakostikova (approved):

katerynakostikova:
Fix conflicts and we'll merge

Issue #132, opened 2023-08-28
Create Talent Sign up first step form


savinaDaria:
![Image](https://user-images.githubusercontent.com/58756139/263674867-c0cbe99b-a091-405a-b12a-cfbd46103281.png)

- [x] Add bundle sign-up
- [x] Add to bundle components **step1** form
- [x] Add Store at the root
- [x] On step1 form **Next button** data should be validated, use **useAppForm()** hook for this
- [x] After validation save data in Store 

User story:
As a Talent, I want to be able to fill in my profile information, so that Employers can easily access this information via search and contact me if our expectations match.

Acceptance criteria:
- Step 1 of Talent registration "Profile" must have input fields: Profile name, Salary expectations, Job title, Experience, Current Location, Employment type, Briefly tell employers about your experience
- User should be able to see the current step number on the left side of the screen
- The registration step “STEP 01/05” is placed in the upper right corner on the same line with the title “Profile”
- All of this fields must be required and have “*”
- The Profile name, Salary expectations, Job title, Current Location, Briefly tell employers about your experience - should have a placeholders as in Fields[ validation document](https://docs.google.com/spreadsheets/d/1OM1GunJPbMa6u0Dn4rm9Cgh32PBfMNOxv588UZYBM_4/edit?usp=sharing)
- Profile name should be between 2-50 chars and accept only A-Z, a-z, (.), ('), (-), spaces
- Salary expectations should be between 1-4 chars and accept only 0-9 values
- Job title should be a dropdown with (TBD) values
- Year of Experience should be a range input with values from No(0 years) to 5+ years with a step of a 0,5 year
- Current Location should be a dropdown with a list of countries
- Employment type	field should have checkboxes: Full time, Part-time, Freelance (projects), Remotely, Relocation to another country
- Brief description about your experience field should be between 10-250 chars and accept only A-Z, a-z, 0-9, special chars, spaces
- Talent should be able to go to the next step by clicking on the “Next” button
- Talent's information should be saved in the fields when switching between steps

nikita-remeslov:
@ed9ward 

> Salary expectations should have a currency dropdown near its input field to select one of the provided currencies: $, €, ₴ (USD, EUR, UAH)

Where did you find that? I can't find this information anywhere!

PR #144, opened 2023-08-28
bt-72: Add expectation salary type(UAH, USD, EUR)

lavandosya:
I've added to user_details table enum with (UAH, USD, EUR) fields for salary_expectation_type that comes from talent registration flow.

### DIFF
--- a/backend/src/common/enums/currency-type.enum.ts
+++ b/backend/src/common/enums/currency-type.enum.ts
@@ -1,0 +1,7 @@
+const CurrencyType = {
+    UAH: 'UAH',
+    USD: 'USD',
+    EUR: 'EUR',
+} as const;
+
+export { CurrencyType };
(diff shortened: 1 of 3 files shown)


backend/src/migrations/20230828154649_update_user_details_table.ts
```diff
@@ -0,0 +1,26 @@
+import { type Knex } from 'knex';
+
+import { CurrencyType } from '~/common/enums/enums.js';
```
MaksymGavatuk, line 3: For migration don't use imported enums because someone will change this in future and we will have wrong migration

MaksymGavatuk (changes requested):

backend/src/common/enums/currency-type.enum.ts
```diff
@@ -0,0 +1,7 @@
+const CurrencyType = {
+    UAH: 'UAH',
+    USD: 'USD',
+    EUR: 'EUR',
+} as const;
+
+export { CurrencyType };
```
nikita-remeslov, line 7: why do we use 3 currencies? 

I think we need just 1

lavandosya: I just saw this in [Create Talent Sign up first step form](https://github.com/BinaryStudioAcademy/bsa-2023-bsa-talents/issues/132)
Screenshot 2023-08-28 at 19 03 35

PR #154, opened 2023-08-29
bt-71: Create get current user endpoint

### LINKED ISSUE #71: Create get current user endpoint
Implement endpoint for get current user
When user refresh page or close it and after open we need to get token from localStorage and call getCurrentUser endpoint to load current user.

- [x] userRepository (Server)
- [x] add endpoint (Server)
- [x] add user to store (Client)
- [x] load user when page loaded (Client)

MChornomaz:


backend/src/bundles/auth/auth.controller.ts
```diff
@@ -135,6 +141,64 @@ class AuthController extends ControllerBase {
...
+                payload: user,
+            };
+        } catch (error) {
+            throw new HttpError({
+                status: HttpCode.NOT_FOUND,
+                message: 'No user found for provided token',
+                cause: error,
+            });
+        }
+    }
```
anton-otroshchenko: is try catch necessary here?
You could have done this instead:
```
if(!user){
throw new HttpError({
                status: HttpCode.NOT_FOUND,
                message: 'No user found for provided token',
            });
}
```

anton-otroshchenko (changes requested):

anton-otroshchenko (approved):

backend/src/bundles/auth/auth.service.ts
```diff
@@ -80,6 +80,12 @@ class AuthService {
 
         return user.toObject();
     }
+
+    public async findByToken(
+        token: string,
+    ): Promise<UserFindResponseDto | undefined> {
+        return await this.userService.findByToken(token);
```
MaksymGavatuk: You don't need return await (http://hassansin.github.io/Why-return-await-Is-a-Bad-Idea)

backend/src/bundles/auth/auth.service.ts
```diff
@@ -80,6 +80,12 @@ class AuthService {
 
         return user.toObject();
     }
+
+    public async findByToken(
+        token: string,
+    ): Promise<UserFindResponseDto | undefined> {
```
MaksymGavatuk: Better use UserFindResponseDto | null return type

backend/src/bundles/users/user.repository.ts
```diff
@@ -17,6 +18,18 @@ class UserRepository implements Repository {
         return user ? UserEntity.initialize(user) : undefined;
     }
 
+    public async findByToken(
+        tokenString: string,
+    ): Promise<UserEntity | undefined> {
```
MaksymGavatuk: Better use UserEntity | null return type

backend/src/bundles/users/user.repository.ts
```diff
@@ -17,6 +18,18 @@ class UserRepository implements Repository {
...
+        const decodedToken = await token.decode(tokenString);
+
+        const user = await this.userModel
+            .query()
+            .findOne({ id: decodedToken.payload.id });
+
+        return user ? UserEntity.initialize(user) : undefined;
```
MaksymGavatuk:
```suggestion
       if (!user) {
            return null;
       }

        return UserEntity.initialize(user);
```

backend/src/bundles/users/user.service.ts
```diff
@@ -47,6 +47,13 @@ class UserService implements Service {
         };
     }
 
+    public async findByToken(
+        token: string,
+    ): Promise<UserFindResponseDto | undefined> {
```
MaksymGavatuk: Better use UserFindResponseDto | null return type

backend/src/bundles/users/user.service.ts
```diff
@@ -47,6 +47,13 @@ class UserService implements Service {
...
+        const userData = await this.userRepository.findByToken(token);
+        return userData ? userData.toObject() : undefined;
```
MaksymGavatuk: The same

frontend/src/bundles/auth/auth-api.ts
```diff
@@ -35,6 +36,18 @@ class AuthApi extends HttpApiBase {
...
+    public async getByToken(): Promise<UserFindResponseDto> {
+        const response = await this.load(
+            this.getFullEndpoint(AuthApiPath.CURRENT_USER, {}),
+            {
+                method: 'GET',
+                contentType: ContentType.JSON,
+                hasAuth: true,
+            },
+        );
+        return await response.json<UserFindResponseDto>();
```
MaksymGavatuk: You don't need return await (http://hassansin.github.io/Why-return-await-Is-a-Bad-Idea)

MaksymGavatuk (changes requested):

backend/src/bundles/auth/auth.controller.ts
```diff
@@ -135,6 +141,63 @@ class AuthController extends ControllerBase {
...
+
+    private async getCurrentUser(
+        options: ApiHandlerOptions,
+    ): Promise<ApiHandlerResponse> {
+        const [, token] = options.headers.authorization?.split(' ') ?? [];
+
+        if (!token) {
+            throw new HttpError({
+                status: HttpCode.UNAUTHORIZED,
+                message: 'Authorization token is missing',
```
nikita-remeslov: ErrorMessage enum

backend/src/bundles/auth/auth.controller.ts
```diff
@@ -135,6 +141,63 @@ class AuthController extends ControllerBase {
...
+            });
+        }
+
+        const user = await this.authService.findByToken(token);
+
+        if (!user) {
+            throw new HttpError({
+                status: HttpCode.NOT_FOUND,
+                message: 'No user found for provided token',
```
nikita-remeslov: Error message enum

backend/src/bundles/auth/auth.controller.ts
```diff
@@ -135,6 +141,63 @@ class AuthController extends ControllerBase {
...
+            });
+        }
```
nikita-remeslov: looks like all this logic should be in the service layer

backend/src/bundles/auth/auth.service.ts
```diff
@@ -80,6 +80,13 @@ class AuthService {
 
         return user.toObject();
     }
+
+    public async findByToken(
```
nikita-remeslov:
```suggestion
    public async getCurrentUser(
```

nikita-remeslov (changes requested):

backend/src/bundles/auth/auth.controller.ts
MChornomaz: I think only this part should be moved 
if (!user) {
            throw new HttpError({
                status: HttpCode.NOT_FOUND,
                message: 'No user found for the provided token',
            });
        }
        
        if (!token)   - is quite logically to leave here

nikita-remeslov: you can handle it all in the service. It will be much cleaner. Controller will return data and Service will be handle all magic

nikita-remeslov (approved):

PR #158, opened 2023-08-29
bt-136: Convert token service to token package

### LINKED ISSUE #136: Convert Token Service to Token Package
Currently token service is defined here:
`backend/src/common/services`

We need to redefine it as a package:
`backend/src/common/token`

Follow style and structure from logger package:
`backend/src/common/logger`

After conversion, combine all packages into `package` dir with root file packages.ts:

- [x]  Convert token service to package
- [x]  Consolidate packages into package dir. `backend/src/common/packages`

vladjsp:


vladjsp:
What has been done:
1. Token service converted into package and moved into package folder.
2. Token interface and class was renamed to remove word 'service' from it. `TokenServiceBase -> TokenBase, TokenService -> Token`.
3. I left the word "Service" in the instance of TokenBase, because I'm afraid to confuse other developers if we will have variables with the same name "token", and this name seems to match the style of the project:
```
type AuthOptions = {
    services: {
        userService: UserService;
        tokenService: Token;
    };
};
```
I am not sure about including controller, database and config into  the packages folder. Will it be more clear and readable than it is now?

backend/src/common/packages/token/token.ts
```diff
@@ -0,0 +1,5 @@
+import { TokenBase } from './token.package.js';
+
+const tokenService = new TokenBase();
```
nikita-remeslov:
```suggestion
const token = new TokenBase();
```

nikita-remeslov (changes requested):
All that is related to the package move to the package folder.

Next time ping someone in the PR if you want to get a quick answer

vladjsp:
Please review this PR in detail. It works fine for me, but I've struggled a lot with conflicts and other errors due to renaming and moving files. This can break the backend and block work for many people, so check it very carefully.

anton-otroshchenko (approved):

ptarho (approved):

backend/src/common/packages/packages.ts
```diff
@@ -0,0 +1,4 @@
+export { config } from './config/config.js';
+export { database } from './database/database.js';
+export { logger } from './logger/logger.js';
+export { token } from './token/token.js';
```
nikita-remeslov, line 5: add controller

nikita-remeslov (changes requested):

backend/src/common/packages/packages.ts
vladjsp: But controller does not exports instance like other 4 packages. Here is what controller exports from controller.ts:
```
export { ControllerBase } from './controller-base.package.js';
export { ControllerHooks } from './enums/enum.js';
export {
    type ApiHandlerOptions,
    type ApiHandlerResponse,
} from './types/types.js';
```
Should I just add into the packages.ts file:
```
...4 previous,
export {ControllerBase, ControllerHooks} from './controller/controller.js'
```
?

nikita-remeslov:
Conflicts

nikita-remeslov (approved):

PR #177, opened 2023-08-30
bt-170: Add local font files

### LINKED ISSUE #170: Add font to assets

ptarho:
I have downloaded font with styles(font-weight) that are yoused in our figma.
Imported all font files via @font-face prop to fonts.scss and then to our main style.scss
After project is build all font files is also included in it in the next way:
build
Screenshot_1


anton-otroshchenko (approved):

frontend/src/assets/fonts/fonts.scss
```diff
@@ -0,0 +1,27 @@
+@font-face {
+    font-weight: 400;
+    font-family: Inter;
+    font-style: normal;
+    src: url("./inter-regular.ttf") format("truetype");
+}
```
anogire, line 6: yes, after using lowercase letters everything should be ok

anogire (approved):

vladjsp (approved):

nikita-remeslov (commented):
I think we should use material theme for it - https://mui.com/material-ui/customization/typography/#self-hosted-fonts

nikita-remeslov (approved):

PR #180, opened 2023-08-30
bt-176: Add placeholder to select component

### LINKED ISSUE #176: Add placeholder to select component

MChornomaz:


vladjsp:
I still don't see the placeholder on select. It appears when I click the select element but not in default (untouched) state.
`<Select placeholder='Options' name='email' control={control}  options={[ { value:'one', label:'one' } ]} label='label'/>`
select1
select3

MChornomaz:
Hmm, need to check one thing. I need 20 minutes

MChornomaz:
To fix this behavior it is necessary to pass to useAppForm default value of this select as any value but an empty string, it is an MI select problem. I suggest passing" ". Then everything works correctly

vladjsp (approved):

frontend/src/bundles/common/components/select/select.tsx
```diff
@@ -17,6 +17,7 @@ type Properties<T extends FieldValues> = {
     control: Control<T, null>;
     name: FieldPath<T>;
     options: { value: string | number; label: string }[];
+    placeholder: string;
```
nikita-remeslov:
```suggestion
    placeholder?: string;
```

frontend/src/bundles/common/components/select/select.tsx
```diff
@@ -35,14 +36,15 @@ const Select = <T extends FieldValues>({
...
+    placeholder,
 }: Properties<T>): JSX.Element => {
     const firstElementIndex = 0;
     const { field } = useFormController({
         name,
         control,
         defaultValue: (multiple
             ? [options[firstElementIndex].value]
-            : options[firstElementIndex].value) as PathValue<T, Path<T>>,
+            : ' ') as PathValue<T, Path<T>>,
```
nikita-remeslov: why do you need this change?

frontend/src/bundles/common/components/select/select.tsx
```diff
@@ -35,14 +36,15 @@ const Select = <T extends FieldValues>({
     multiple,
     hasError,
     isDisabled,
+    placeholder,
```
nikita-remeslov:
```suggestion
    placeholder = "Placeholder",
```

nikita-remeslov (changes requested):

frontend/src/bundles/common/components/select/select.tsx
MChornomaz: They were not really necessary, fixed

nikita-remeslov (approved):

PR #197, opened 2023-08-31
bt-124: Chat message list component

### LINKED ISSUE #124: Create chat message list component
Image

ptarho:
I have added chat avatar (will probabl replace it from other developer branch), chat-list and chat-message components.
Currently we don`t save user info in a store, so I mocked userId variable to apply different styles to chat-message. 
Also changed default scrollbar inside chat-list. 
That is how we would use this components and how they looks now:
usage_code
Message list take array of messages in the next format:
messages_array
Chat header and footer will be replaced by another developer after they will be implemented.
Screenshot_1
Screenshot_2


frontend/src/bundles/common/components/avatar/avatar.tsx
```diff
@@ -0,0 +1,33 @@
...
+    size = 'small',
+    userFullName,
+}) => {
+    const avatarClasses = getValidClassNames(styles.avatar, styles[size]);
+
+    const [userName, userSurname] = userFullName.split(' ');
+    const firstLetterIndex = 0;
+    const userNameAbbreviation = `${userName[
+        firstLetterIndex
+    ].toUpperCase()}${userSurname[firstLetterIndex].toUpperCase()}`;
```
JemiJe: if we have only for ex. without space 'name', it will be error. after `userFullName.split(' ');`, `userSurname` will be `undefined`, then we trying `userSurname[0]`, may be better take one letter from username string.

frontend/src/bundles/common/components/avatar/avatar.tsx
```diff
@@ -0,0 +1,33 @@
...
+    return url ? (
+        <img className={avatarClasses} src={url} alt="User avatar" />
```
JemiJe: we have already alt User avatar for header, maybe better alt will be `User chat avatar`?

frontend/src/bundles/common/components/avatar/styles.module.scss
```diff
@@ -0,0 +1,25 @@
+.avatar {
+    img {
+        width: 100%;
+        height: 100%;
+    }
+
+    display: flex;
+    justify-content: center;
+    align-items: center;
```
JemiJe: I have similar task with flex, and reviews required to use props of Grid component instead of styles. but maybe еhat is unnecessary

frontend/src/bundles/common/components/avatar/avatar.tsx
```diff
@@ -0,0 +1,36 @@
...
+}) => {
+    const avatarClasses = getValidClassNames(styles.avatar, styles[size]);
+
+    const [userName, userSurname] = userFullName.split(' ');
+    const firstLetterIndex = 0;
+    const userNameAbbreviation = userSurname
+        ? `${userName[firstLetterIndex].toUpperCase()}${userSurname[
+              firstLetterIndex
+          ].toUpperCase()}`
+        : `${userName[firstLetterIndex].toUpperCase()}`;
```
MaksymGavatuk, line 25: I think we need to create some general helpers for this logic. Because I already saw it somewhere

frontend/src/bundles/common/components/avatar/avatar.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+const Avatar: React.FC<Properties> = ({
+    url,
+    size = 'small',
+    userFullName,
+}) => {
+    const avatarClasses = getValidClassNames(styles.avatar, styles[size]);
+
+    const [userName, userSurname] = userFullName.split(' ');
+    const firstLetterIndex = 0;
```
MaksymGavatuk: Move it to some constant

frontend/src/bundles/common/components/avatar/avatar.tsx
MaksymGavatuk: I think we need to create some general helpers for this logic. Because I already saw it somewhere

MaksymGavatuk (changes requested):

frontend/src/bundles/common/components/avatar/avatar.tsx
vladjsp: fixed this in commit 088ac8d450436eb00333a7a1a350a9c30d119b23 in [task/bt-126-create-chat-header-and-send-message](https://github.com/BinaryStudioAcademy/bsa-2023-bsa-talents/pull/199)

vladjsp:
do we care about firefox users? left FF, right Chrome
messages-list

ptarho:
> do we care about firefox users? left FF, right Chrome messages-list

Fixed styles for it

JemiJe:
I my opinion, we should remove unnecessary words in avatar, looks not very pretty. Inital design in figma was minimalistic and light, but these words adds some weight to perceive, as for me. also, that feature wasn't required. 
What are the authors of initial design think?
compare
Screenshot 2023-09-01 200412
Screenshot 2023-09-01 200458

JemiJe:
merged with new Avatar, changed class names with new ones according to new requirements, changed scroll bar style as we have on mockup (maybe should be created mixin for all project)
Screenshot 2023-09-07 114907

nikita-remeslov:
merge development branch please

shared/src/helpers/user/get-avatar-initials.ts
```diff
@@ -1,9 +1,12 @@
 const getAvatarInitials = (fullName: string): string => {
     const [name, secondName] = fullName.split(' ');
     const firstCharIndex = 0;
-    return `${name[firstCharIndex].toUpperCase()}${secondName[
-        firstCharIndex
-    ].toUpperCase()}`;
+
```
nikita-remeslov, line 4: I think we have those changes in the few PRs.

nikita-remeslov (approved):

anton-otroshchenko (approved):

PR #184, opened 2023-08-30
bt-159: add sign up flow

illia-dobro:
The validation schema works; data from the fields is being sent. 

signUpss

PR #198, opened 2023-08-31
bt-193: Add wrapper for sign in/up screens

### LINKED ISSUE #193: Add wrapper for sign in/up screens
Currently we have only form UI for sign in/sign up, since we already have App logo we can complete the UI for these screens
Design link: [https://www.figma.com/file/Yd7USgP7PLvA0W794jHXWu/BSA_recruting?type=design&node-id=4266-1041&mode=design&t=fyH0m4tlKerCXYlp-0](https://www.figma.com/file/Yd7USgP7PLvA0W794jHXWu/BSA_recruting?type=design&node-id=4266-1041&mode=design&t=fyH0m4tlKerCXYlp-0)

Image

likeri29:
login
sign up

Changed a bit styles in the sign-in and sign-up screens. 

Had to disable two linter rules because it doesn't give opportunity to use `require`, and I need it for image.


nikita-remeslov (changes requested):
we should add a checkbox to understand, who wants to sign up

image

mobile/src/bundles/auth/components/auth-wrapper/auth-wrapper.tsx
```diff
@@ -0,0 +1,35 @@
...
+import { globalStyles } from '~/bundles/common/styles/styles';
+
+import { styles } from './styles';
+
+type Properties = {
+    children: React.ReactNode;
+};
+
+// eslint-disable-next-line unicorn/prefer-module, @typescript-eslint/no-var-requires
+const logoImg = require('~/assets/images/logo.png');
```
nikita-remeslov: I am not sure about `require`, we should use modules

mobile/src/navigations/root/root.tsx
```diff
@@ -28,7 +28,7 @@ type Properties = {
 
 const Root: React.FC<Properties> = ({
     isSignedIn = false,
-    isProfileComplete = true,
+    isProfileComplete = false,
     role = 'talent',
 }) => {
     // prettier-ignore
```
nikita-remeslov: why do we ignore prettier?

likeri29:
> we should add a checkbox to understand, who wants to sign up
> 
> image

Illia is doing it in his [PR](https://github.com/BinaryStudioAcademy/bsa-2023-bsa-talents/pull/184), I just created wrapper.

nikita-remeslov (changes requested):

mobile/src/navigations/root/root.tsx
likeri29: Because we have a conflict between linter and prettier, we've decided temporarily to place `prettier-ignore` there

mobile/src/bundles/auth/components/auth-wrapper/auth-wrapper.tsx
likeri29: Am I allowed to create index.d.ts file in the mobile src folder?

mobile/src/navigations/root/root.tsx
katerynakostikova: There are multiple `? :` checks and prettier removes brackets and there is linter error

MaksymGavatuk: You can move this logic to renderStackScreen function for example and use early return to prevent multiple ? :

MaksymGavatuk: And in this case it would be more human readable as for me

MaksymGavatuk: @katerynakostikova what do you think about this idea ?

katerynakostikova: Let's do like this

likeri29: Remake it in this PR?

katerynakostikova: yes

likeri29: Done

mobile/src/assets/images/logo.png
katerynakostikova, line 1: let's call it logo with name

mobile/src/bundles/auth/components/sign-up-form/styles.ts
```diff
@@ -0,0 +1,25 @@
+import { StyleSheet } from 'react-native';
+
+import { Color, FontFamily } from '~/bundles/common/enums/enums';
+
+const styles = StyleSheet.create({
+    container: {
+        gap: 10,
+        backgroundColor: '#FFFFFF',
+        width: 345,
```
katerynakostikova: let's avoid this magic numbers

mobile/src/bundles/auth/screens/auth.tsx
```diff
@@ -48,10 +48,18 @@ const Auth: React.FC = () => {
...
+                    </AuthWrapper>
+                );
             }
             case AuthScreenName.SIGN_UP: {
-                return <SignUpForm onSubmit={handleSignUpSubmit} />;
+                return (
+                    <AuthWrapper>
+                        <SignUpForm onSubmit={handleSignUpSubmit} />
+                    </AuthWrapper>
+                );
```
katerynakostikova: You can add Auth Wrapper below instead of duplicating the code

mobile/src/navigations/root/root.tsx
```diff
@@ -21,32 +24,57 @@ const screenOptions: NativeStackNavigationOptions = {
...
+                        : EmployerBottomTabNavigator
+                }
+            />
+        );
+    }
+
+    if (isProfileComplete) {
+        return (
+            <RootStack.Screen
+                name={RootScreenName.ONBOARDING_ROOT_ROUTE}
```
katerynakostikova: You can move this function into Root and use values directly

katerynakostikova (approved):

nikita-remeslov (approved):

PR #249, opened 2023-09-05
bt-248:  Swagger not attaching token

### LINKED ISSUE #248: [BUG] Swagger not attaching token
Need to update markup to attach token in swagger.

auth.controller.getCurrentUser
user.controller.findAll

JosuerBague:
Changed annotation. Now token is attached when authorized in swagger.

### DIFF
--- a/backend/src/bundles/users/user.controller.ts
+++ b/backend/src/bundles/users/user.controller.ts
@@ -10,6 +10,11 @@ import { UsersApiPath } from './enums/enums.js';
 /**
  * @swagger
  * components:
+ *    securitySchemes:
+ *     bearerAuth:
+ *       type: http
+ *       scheme: bearer
+ *       bearerFormat: JWT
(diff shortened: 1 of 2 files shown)


Monyster (dismissed):

backend/src/bundles/users/user.controller.ts
```diff
@@ -10,6 +10,11 @@ import { UsersApiPath } from './enums/enums.js';
 /**
  * @swagger
  * components:
+ *    securitySchemes:
```
ptarho: I am not sure that we have to duplicate it here, this schema was already declared in auth controller

JosuerBague: You are correct, there was no need for it

MChornomaz (approved):

nikita-remeslov (approved):

PR #253, opened 2023-09-05
bt-243: Fix arrow and colors in Select component

### LINKED ISSUE #243: Fix arrow and colors in Select component

MChornomaz:


frontend/src/bundles/common/components/select/select.tsx
```diff
@@ -39,6 +42,8 @@ const Select = <T extends FieldValues>({
     placeholder = 'Placeholder',
 }: Properties<T>): JSX.Element => {
     const firstElementIndex = 0;
+    const secondElementIndex = 1;
```
savinaDaria: Maybe this should be in capital letters, [B22](https://github.com/BinaryStudioAcademy/quality-criteria/blob/production/src/javascript.md)

savinaDaria (approved):
I have tested it on First onboarding step,  works. First value is selected, second with Option as placeholder - not selected,
Also click on arrow now works
Usage:
```
<Select
    control={control}
    name={'jobTitle'}
    options={jobTitleOptions}
    placeholder="Option"
/>
  ```
image

File: frontend/src/bundles/common/components/select/select.tsx (hunk 1 above)
Comment (MChornomaz) [reply frontend/src/bundles/common/components/select/select.tsx]: Thanks

File: frontend/src/bundles/common/components/select/select.tsx [hunk 2]
```diff
@@ -38,15 +41,59 @@ const Select = <T extends FieldValues>({
...
+                    .slice(SECOND_ELEMENT_INDEX)
+                    .map((value: string | number) => (
+                        <span key={value} className={styles.option}>
+                            {
+                                options.find((item) => item.value === value)
+                                    ?.label
+                            }
+                            ,&nbsp;
+                        </span>
+                    ));
```
anton-otroshchenko: I think this function is very complicated. Maybe it will be a good idea to put this logic into a different function to make it more readable

anton-otroshchenko (changes requested):

frontend/src/bundles/common/components/select/select.tsx
```diff
@@ -38,15 +41,84 @@ const Select = <T extends FieldValues>({
...
+            }
+        },
+        [
+            options,
+            placeholder,
+            renderPlaceholder,
+            renderSelectedOptions,
+            renderSingleSelectedOption,
+        ],
+    );
```
nikita-remeslov: what the magic? can we simplify it?

nikita-remeslov (changes requested):

frontend/src/bundles/common/components/select/select.tsx
MChornomaz: fixed

anton-otroshchenko (approved):

nikita-remeslov (changes requested):
This is not what we discussed. It treats the symptom, not the problem as a whole.

nikita-remeslov (changes requested):
Merge development branch and update props please

nikita-remeslov (approved):

PR #296, opened 2023-09-07
bt-287: add get current user request to mobile

### LINKED ISSUE #287: Add get current user request to mobile
Currently, when the user is signed in/up we still need to sign in again after the app relaunch. Let's add get current user request to fetch user data with stored token. If request failed we need to run logout action and user should be redirected to sign in screen

illia-dobro:


mobile/src/bundles/auth/screens/auth.tsx
```diff
@@ -14,23 +14,22 @@ import {
...
-    const isSignUpScreen = name === AuthScreenName.SIGN_UP;
+
     const isPendingAuth = dataStatus === DataStatus.PENDING;
+
     useEffect(() => {
-        if (isSignUpScreen) {
-            void dispatch(userActions.loadAll());
-        }
-    }, [isSignUpScreen, dispatch]);
+        void dispatch(loadUser());
```
katerynakostikova: It should be dispatched in root, user shouldn't see sign in screen each time he/she opens the app

mobile/src/bundles/auth/auth-api.ts
```diff
@@ -38,6 +39,18 @@ class AuthApi extends HttpApiBase {
         return await response.json<UserSignUpResponseDto>();
     }
 
+    public async getByToken(): Promise<UserFindResponseDto> {
```
katerynakostikova:
```suggestion
    public async getCurrentUser(): Promise<UserFindResponseDto> {
```

mobile/src/bundles/auth/store/actions.ts
```diff
@@ -30,11 +31,20 @@ const signUp = createAsyncThunk<
     }
 });
 
+const loadUser = createAsyncThunk<
```
katerynakostikova:
```suggestion
const loadCurrentUser = createAsyncThunk<
```

mobile/src/navigations/root/root.tsx
```diff
@@ -24,6 +30,21 @@ const screenOptions: NativeStackNavigationOptions = {
...
+
+    const isPendingAuth = dataStatus === DataStatus.PENDING;
+
+    useEffect(() => {
+        void dispatch(loadUser());
+    }, [dispatch]);
+
+    if (isPendingAuth) {
+        return <Overlay isActive={isPendingAuth} />;
+    }
```
katerynakostikova:
```suggestion
    if (isPendingAuth) {
        return null
    }
```

katerynakostikova (approved):

anogire (approved):

PR #396, opened 2023-09-12
bt-330: [BUG] The "Step" in the right upper corner is moved to another line

illia-dobro:
I have found out how to recreate this bug on the emulator. I increase the font size in the phone's settings and then observe it.
max

### DIFF
--- a/mobile/src/navigations/onboarding-navigator/components/header/styles.ts
+++ b/mobile/src/navigations/onboarding-navigator/components/header/styles.ts
@@ -16,8 +16,8 @@ const styles = StyleSheet.create({
         color: '#fff',
     },
     steps: {
-        width: 68,
-        height: 68,
+        width: 65,
+        height: 65,
         borderRadius: 100,
         backgroundColor: '#d6dce8',
         position: 'absolute',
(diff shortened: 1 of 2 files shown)


mobile/src/navigations/onboarding-navigator/components/header/styles.ts
```diff
@@ -16,8 +16,8 @@ const styles = StyleSheet.create({
         color: '#fff',
     },
     steps: {
-        width: 65,
-        height: 65,
+        width: 68,
+        height: 68,
```
katerynakostikova: why changing this?

illia-dobro: To create some space for 'STEP'. These changes are not necessary, but it appears to be very close to breaking without them.

anogire: maybe some margins are better?

illia-dobro: I returned them to their previous values; it seems to work without changing them.

katerynakostikova (approved):

anogire (approved):

PR #427, opened 2023-09-14
bt-393: add loader on application start to get current user action

### LINKED ISSUE #393: Add loader on application start. To get current user action

mskmee:
image


mobile/src/navigations/root/root.tsx
```diff
@@ -31,18 +31,10 @@ const Root: React.FC = () => {
     const { isProfileComplete, role } = userData ?? {};
     const dispatch = useAppDispatch();
 
-    const { dataStatus } = useAppSelector(({ auth }) => auth);
-
-    const isPendingAuth = dataStatus === DataStatus.PENDING;
-
```
katerynakostikova: it should stay here, user shouldn't see sign in screen each time app loads
here we should return Loader (white screen with spinner) 
in auth form we'll return overlay

katerynakostikova (changes requested):

mskmee:
image

mobile/src/bundles/common/components/loader/loader.tsx
```diff
@@ -8,9 +8,9 @@ import { Color } from '~/bundles/common/enums/enums';
 import { globalStyles } from '~/bundles/common/styles/styles';
 
 type LoaderProperties = {
-    size: 'small' | 'large';
+    size?: 'small' | 'large';
```
katerynakostikova: enum

katerynakostikova (approved):

anogire (approved):

PR #465, opened 2023-09-15
bt-423: Add dropdown for sorting search reasults

### LINKED ISSUE #423: Add Dropdown component for sorting the search result
Acceptance criteria:

1. Should apply requirements for common dropdown component
2. Should have options
- sort by salary in ascending order 
- sort by salary in descending order 
- sort by experience in ascending order 
- sort by experience in descending order 
- sort by newest 
- sort by oldest 

sort

vladjsp:
sorting
sorting22


MChornomaz:
There is a small but annoying error in the console. It is connected with the mapping of candidates. It is not really a part of this PR, but it should be better, to make small fixes to mockCandidates to solve it.
image

frontend/src/bundles/employers/enums/enums.ts
```diff
@@ -3,6 +3,7 @@ export { BsaCharacteristics } from './bsa-characteristics.enum.js';
 export { BsaProject } from './bsa-project.enum.js';
 export { CheckboxesFields } from './checkboxes-fields-name.enum.js';
 export { ExperienceYears } from './experience-years.enum.js';
+export { SortingOptions } from './sorting-options.enum.js';
```
MChornomaz: I think It is better to use this enum, change its labels to those, that are necessary in UI and map through them, we will have all values logic in 1 place, and it will not be necessary to make changes in 2 places if something changes, also debugging will be easier
image

MChornomaz (changes requested):

MChornomaz (approved):

frontend/src/bundles/employers/components/sorting-dropdown/sorting-dropdown.tsx
```diff
@@ -0,0 +1,42 @@
+import { InputAdornment } from '@mui/material';
```
nikita-remeslov: why do we need this component?

frontend/src/bundles/employers/components/sorting-dropdown/sorting-dropdown.tsx
```diff
@@ -0,0 +1,42 @@
...
+import {
+    type Control,
+    type FieldPath,
+    type UseFormReset,
+} from 'react-hook-form';
```
nikita-remeslov, line 5: common bundle

nikita-remeslov (changes requested):

frontend/src/bundles/employers/components/sorting-dropdown/sorting-dropdown.tsx
vladjsp: moved InputAdornment into common Select component as optional props. Since this field has no label, I decided to place this adornment to clarify the purpose of the field. Or we can add 'Sort by ...' to all options, but that, a lot more text and sorting field will be left unlabeled.

frontend/src/bundles/employers/components/sorting-dropdown/sorting-dropdown.tsx
vladjsp: this is the common way to import rhf types in our project. Every component that needs these types (not just the common components) imports them directly. We currently do not have an export bundle for these types. Should we create it and change all the components?

nikita-remeslov: sorry it is types!

nikita-remeslov (approved):

PR #480, opened 2023-09-15
bt-476: Remove Back button from Step 1

### LINKED ISSUE #476: Remove Back button from Step 1
There's no point in this button since user can't go Back from the first step. Next button should be placed in the same place and remain untouched.

illia-dobro:
as

### DIFF
--- a/mobile/src/bundles/talent/components/profile-form/styles.ts
+++ b/mobile/src/bundles/talent/components/profile-form/styles.ts
@@ -5,9 +5,6 @@ const styles = StyleSheet.create({
         backgroundColor: '#FFFFFF',
         flexGrow: 1,
     },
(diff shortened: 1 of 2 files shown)


mskmee (approved):

mobile/src/bundles/talent/components/profile-form/styles.ts
```diff
@@ -5,6 +5,9 @@ const styles = StyleSheet.create({
         backgroundColor: '#FFFFFF',
         flexGrow: 1,
     },
+    submitButton: {
+        marginLeft: 100,
+    },
```
katerynakostikova: let's remove this margin

katerynakostikova (approved):

PR #492, opened 2023-09-17
bt-397: Add transition to elements for Onboarding Flow

### LINKED ISSUE #397: Add transition to elements for Onboarding Flow
When the user hovers over the elements there should be a visual effect
Also, it is necessary to add a transition for states updates
https://developer.mozilla.org/en-US/docs/Web/CSS/transition

Monyster:
Closes #397 

This blue line changes with the transition at each step
image


frontend/src/bundles/common/components/select/styles.module.scss
```diff
@@ -39,9 +46,9 @@
     background-color: transparent;
 }
 
-.container .input:hover {
-    border-color: var(--botticelli);
-}
+// .container .input:hover {
+//     border-color: var(--st-tropaz);
+// }
```
MaksymGavatuk: Remove comments if we don't need this

MaksymGavatuk (changes requested):

nikita-remeslov (changes requested):
in the design blue line in one place. we just need to add a transition on elements state (hover, error, etc.)

Monyster:
@nikita-remeslov I can remove it, but as I see on other sites, it should be a moving underline or progress bar.
I can change it to the progress bar style.

If you look at the design, you'll see that this is not a decorative underline for the step name.
image

I suggest to do as shown next
image
OR
image

frontend/src/bundles/talent-onboarding/constants/constants.ts
```diff
@@ -3,6 +3,10 @@ import { StepsRoute } from '~/bundles/talent-onboarding/enums/enums.js';
 const FIRST_ELEMENT = 0;
 const LAST_TWO_CHARS = -2;
 const STEP_ONE = 1;
+const STEP_TWO = 2;
+const STEP_THREE = 3;
+const STEP_FOUR = 4;
+const STEP_FIVE = 5;
```
nikita-remeslov: should be enum

nikita-remeslov (changes requested):

frontend/src/bundles/talent-onboarding/constants/constants.ts
```diff
@@ -2,7 +2,14 @@ import { StepsRoute } from '~/bundles/talent-onboarding/enums/enums.js';
...
 const FIRST_ELEMENT = 0;
 const LAST_TWO_CHARS = -2;
-const STEP_ONE = 1;
+const StepsList = {
+    ONE: 1,
+    TWO: 2,
+    THREE: 3,
+    FOUR: 4,
+    FIVE: 5,
+};
```
nikita-remeslov: check how we create enums

nikita-remeslov (changes requested):

frontend/src/bundles/talent-onboarding/constants/constants.ts
Monyster: as const

nikita-remeslov:
conflicts

nikita-remeslov (approved):

PR #569, opened 2023-09-20
bt-541: replace candidate screen with real data and connect filters

### LINKED ISSUE #541: Replace candidate screen with real data and connect filters
Image

mskmee:
Filters and cadidate taking data from backend.
It's because we didnt have bades with names from LMS.
image
image
image
image
image


mobile/src/bundles/employer/helpers/transform-candidate-filter-form-to-query/transform-candidate-filter-form-to-query.ts
```diff
@@ -0,0 +1,35 @@
...
+            result += `&englishLevel=${value}`;
+        }
+    }
+    if (employmentType?.length) {
+        for (const value of employmentType) {
+            result += `&employmentType=${value}`;
+        }
+    }
+
+    for (const [key, value] of Object.entries(multiSelectedData)) {
```
mskmee, line 51: thinking about another helper to simplify thoose for of, but dont knwo is it a good idea to create helper for helper?)

mobile/src/bundles/employer/components/candidate-card/candidate-card.tsx
```diff
@@ -9,26 +9,44 @@ import {
...
+const getBadgeById = (data: BadgesItem[], id: string): BadgesItem => {
+    return data.find((value) => value.id === id) as BadgesItem;
+};
+
+const getHardSkillByValue = (
+    data: FormattedHardSkillsItem[],
+    id: string,
+): FormattedHardSkillsItem => {
+    return data.find(({ value }) => value === id) as FormattedHardSkillsItem;
+};
```
katerynakostikova: can be moved to helpers

mobile/src/bundles/employer/employer-api.ts
```diff
@@ -0,0 +1,46 @@
...
+        const response = await this.load(
+            this.getFullEndpoint(UserDetailsApiPath.ROOT, payload, {}),
+            {
+                method: 'GET',
+                contentType: ContentType.JSON,
+                hasAuth: true,
+            },
+        );
+        return await response.json<UserDetailsResponseDto[]>();
+    }
```
katerynakostikova: I think it can be one request, filters will depend on query params

mskmee: Than we will always run create query string function before make request, thougt that it's unnesseray. 
Combine to one request?

katerynakostikova: It's ok

katerynakostikova: Let's combine

katerynakostikova: It's a common thing to have for example default query params, so it should be the same

katerynakostikova (approved):

katerynakostikova:
Conflicts

anogire (approved):

PR #587, opened 2023-09-21
bt-528: split deny and approve

### LINKED ISSUE #528: Implement deny&approve backend logic
Implement backend logic for deny&approve action

Image

Also need to save reason

Image

Approve:

Image

savinaDaria:
Logic for deny & approve was splitted:
image
image


shared/src/bundles/user-details/enums/user-details-api-path.enum.ts
```diff
@@ -1,7 +1,8 @@
 const UserDetailsApiPath = {
     ROOT: '',
     $ID: '/:userId',
-    APPROVE: '/approve',
+    APPROVE: '/approve/:userId',
+    DENY: '/deny/:userId',
```
MaksymGavatuk: Change to /:userId/approve, because first we specify to which user we need to do action and after that action. It is more clear

MaksymGavatuk (changes requested):

nikita-remeslov (approved):

PR #618, opened 2023-09-21
bt-607: Talent onboarding screen change from step3 to step4

anogire:


anogire:
I had to change the frontend as well, but I tried not to break anything.
Please take a look at it.

nikita-remeslov (approved):

nikita-remeslov (changes requested):
we decided to map data form backend to data that we need on frontend

anogire:
the new [PR](https://github.com/BinaryStudioAcademy/bsa-2023-bsa-talents/pull/646)

PR #660, opened 2023-09-24
bt-599: Make changes to the code according to B20, B22, and B23 rules in quality criteria

### LINKED ISSUE #599: Make changes to the code according to B20, B22, and B23 rules in quality criteria
Go through the whole project and fix these and similar issues

Image

Image

Image

nika-sirak:


nikita-remeslov:
we should update it the migration constants too. 
image
also column name and table name should be enums

nikita-remeslov (approved):

Issue #336, opened 2023-09-08
Create endpoint to load userData from LMS


MaksymGavatuk:
Our system pull data from LMS about students, badges, hard skills
You need to implement endpoint for this. We call this endpoint while we use onboarding process to store some data and set some data to onboarding fields.
~~prevent to sign up if email of new user is missing on LMS server~~ will be in task #691

ed9ward:
Tested

PR #717, opened 2023-09-26
bt-691: Validate talent email does it exist on LMS

### LINKED ISSUE #691: Validate talent email does it exist on LMS
prevent to sign up if user typed email which doesn't exist on LMS

When a user enters an email that doesn't exist on LMS and clicks sign up the error message "Please, use your LMS email" (or something like this) should be displayed

JemiJe:
- it will pass further if talent (not employer) provide email that exists on LMS, otherwise, we get error 400 (Bad Request) and warning message.
- account above it if you test sign up process.
- account also `lmsDataService` class instance for `authService` imported from`users` bundle where it used too and reexported from here from original `lms-data` bundle, so it bind to `users` bunde
- related with #336 
Screenshot 2023-09-26 182637
## algorithm 
```
talent enter all credentials and click 'continue' on sign up page
└── get user LMS data from LMS server by email
    ├── if not found
    │   ├── prevent register
    │   ├── send error 400 
    │   └── show warning message
    └── found on LMS server
        ├── save that data in user_lms_data
        ├── user go to onboarding
        └── further requsting lms data will get data from DB usung /user/:userId/lms-data rote 
```

### DIFF
--- a/backend/src/bundles/auth/auth.service.ts
+++ b/backend/src/bundles/auth/auth.service.ts
@@ -1,3 +1,4 @@
+import { lmsDataService } from '~/bundles/lms-data/lms-data.js';
 import {
     type UserFindResponseDto,
     type UserForgotPasswordRequestDto,
@@ -8,7 +9,7 @@ import {
     type UserSignUpResponseDto,
 } from '~/bundles/users/types/types.js';
 import { type UserService } from '~/bundles/users/user.service.js';
-import { ErrorMessage } from '~/common/enums/enums.js';
+import { ErrorMessage, UserRole } from '~/common/enums/enums.js';
 import { HttpCode, HttpError } from '~/common/http/http.js';
 import { type Encrypt } from '~/common/packages/encrypt/encrypt.js';
 import { token } from '~/common/packages/packages.js';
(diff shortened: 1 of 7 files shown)


backend/src/bundles/auth/auth.service.ts
```diff
@@ -49,8 +56,21 @@ class AuthService {
...
+        if (!dataFromLMS && isUserTalent) {
+            throw new HttpError({
+                message: ErrorMessage.NOT_FOUND_ON_LMS,
+                status: HttpCode.BAD_REQUEST,
+            });
+        }
+
         const user = await this.userService.create(userRequestDto);
 
+        await this.lmsDataService.addUserLMSDataToDB(user.id, dataFromLMS);
```
MaksymGavatuk: As improvement, we can do data fetching from LMS only when role === UserRole.TALENT, because for now we just do unused request for users which have other role

MaksymGavatuk (approved):
Looks good for me, only one comment about fetching LMS data, but not critical

backend/src/bundles/auth/auth.service.ts
JemiJe: really, I will fix it

shared/src/enums/error-message.enum.ts
```diff
@@ -13,7 +13,8 @@ const ErrorMessage = {
     NOT_IMPLEMENTED: 'Not implemented.',
     USER_DETAILS_NOT_FOUND: 'User profile not found',
     TOKEN_INVALID_OR_EXPIRED: 'token invalid or expired.',
-    NOT_FOUND_ON_LMS: 'User not found on LMS server',
+    NOT_FOUND_ON_LMS:
+        'User Not Found: To sign up, please use the same email address you used when registering with Binary Studio Academy',
```
nikita-remeslov:
```suggestion
        'User Not Found! To complete the registration process, please use the same address as when registering at the Binary Studio Academy',
```

nikita-remeslov (approved):
Looks good! just change error message

PR #764, opened 2023-09-27
bt-706: Remove notification on sign-in and sign-up pages

### LINKED ISSUE #706: Remove Fetch when user sign up
Image

vladjsp:
closes https://github.com/BinaryStudioAcademy/bsa-2023-bsa-talents/issues/706

https://github.com/BinaryStudioAcademy/bsa-2023-bsa-talents/assets/118971905/58260f7c-03a7-42fb-adc3-348b8d49a70b


backend/src/common/plugins/authorization/authorization.plugin.ts
```diff
@@ -51,10 +51,7 @@ const authorizationPlugin: FastifyPluginCallback<AuthOptions> = (
         const [, token] = authorization?.split(' ') ?? [];
 
         if (!token) {
-            throw new HttpError({
-                message: ErrorMessage.UNAUTHORIZED_USER,
-                status: HttpCode.UNAUTHORIZED,
-            });
+            return;
```
MaksymGavatuk: Are you sure it the correct way to fix it ?
In this case when we don't have token we will not get info about it in frontend when we do request to not white route, did you test this case ?

MaksymGavatuk (changes requested):

backend/src/common/plugins/authorization/authorization.plugin.ts
vladjsp: I will check it.

vladjsp: You was right. I added checks for auth pages and disabled the error throw only for them.

vladjsp:
In order not to bother users with a big red server error notification when they first visit our web application, my task was to remove this notification for auth pages (sign-in, sign-up), but leave it inside for non-white routes.

anton-otroshchenko (approved):

MaksymGavatuk (approved):
Yes, now it is correct way, can be merged

nikita-remeslov (approved):

